PR #59, opened 2022-04-05
Add support for --output flag to workloads list command

### LINKED ISSUE #12: add `--output json` to `workloads list` command
### Description of problem

I'd like to get a list of workloads in json format for scripting and programmatic parsing.

### Proposed solution

```gherkin
Given I do `tanzu apps workload list --output json`
Then I see output of current information in json format
```

shashwathi:
Add `--output` flag to workloads list command.  This flag supports both yaml and json format

Fixes #12

### Testing done for the PR

```
tanzu apps workload list --output json
 [
  	{
  		"kind": "Workload",
  		"apiVersion": "carto.run/v1alpha1",
  		"metadata": {
  			"name": "petclinic",
  			"namespace": "default",
.....
       }
]
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -156,6 +156,9 @@ jobs:
         echo "##[group]List workload"
           tanzu apps workload list
         echo "##[endgroup]"
+        echo "##[group]List workload with output json"
+          tanzu apps workload list --output json | jq '.[] | select(.metadata.name=="petclinic")'
+        echo "##[endgroup]"
         echo "##[group]Get workload"
           tanzu apps workload get petclinic
         echo "##[endgroup]"
--- a/pkg/cli-runtime/printer/resource.go
+++ b/pkg/cli-runtime/printer/resource.go
@@ -78,17 +78,36 @@ func ExportResource(obj Object, format OutputFormat, scheme *runtime.Scheme) (st
 	return printObject(u, format)
 }
 
-func OutputResource(obj Object, format OutputFormat, scheme *runtime.Scheme) (string, error) {
+func setGVK(obj Object, scheme *runtime.Scheme) (Object, error) {
 	copy := obj.DeepCopyObject().(Object)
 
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -156,6 +156,9 @@ jobs:
         echo "##[group]List workload"
           tanzu apps workload list
         echo "##[endgroup]"
+        echo "##[group]List workload with output json"
+          tanzu apps workload list --output json
```
rashedkvm, line 160: nit: This is good to validate in CI. What do you think about adding a `jsonpath` validation from the output?

Example: ` -o jsonpath='{.status.conditions[?(@.type == "Ready")].status'})`

pkg/cli-runtime/printer/resource.go
```diff
@@ -91,6 +91,22 @@ func OutputResource(obj Object, format OutputFormat, scheme *runtime.Scheme) (st
 	return printObject(copy, format)
 }
 
+func OutputResources(objList []Object, format OutputFormat, scheme *runtime.Scheme) (string, error) {
```
rashedkvm, line 94: Can these two functions be consolidated `OutputResource` & `OutputResources`?

rashedkvm (commented):
PR looks good. Few enhancements in the comments. Something to think about and can be addressed as separate PR, related to the `output` formats we currently have `json` and `yaml`. Should there be `yml` as well?

pkg/cli-runtime/printer/resource.go
warango4: +1

warango4 (approved):
LGTM

.github/workflows/ci.yaml
shashwathi: there is no status blocks so I have updated to check for workload with the name `petclinic`

rashedkvm (approved):

PR #65, opened 2022-04-07
Add support for subPath for source 

### LINKED ISSUE #35: --sub-path parameter supported for `--local-path` and `--git-*`
### Description of problem

* Currently there is not a way for someone to specify a sub-section of a repo or a local project directory to create/update their workload from.
* Additionally, if there is a code change anywhere in the repo, any workload that references the repo will be rebuilt, even if there are no changes in the workload code.

Many GIT ?repositories contain many sub-modules in the same project, so we need to let the pipeline know which sub-path we want to use for the workload creation.

### Proposed solution

```gherkin
Given I'm currently in the root directory of project in A which includes sub-component dirs B and C
When I run "tanzu apps workload create/update/apply my-app --local path . --sub-path B/
Then the entire contents of my project are uploaded and the workload object includes both spec.source.image=source-image-reference and spec.source.subPath=B
```

If an absolute path is provided for the sub-path value (for either `--git*` or `--local-path`), and the sub-path is legit, the plugin should be able to handle that scenario gracefully (meaning - it should work and not error).

### Open Questions:
* can the sub-path be changed after workload creation?
* can the sub-path be removed after workload creation? 
* can you just pass in sub-path in a workload update/apply (for existing workload - with and/or without previously configured sub-path)
* can/should it be possible to remove --sub-path if previously set? (this question is broader than just sub-path
  * if we decide on a pattern for this, we must apply it to other flags (limit-cpu, limit-memory, request-cpu,  request-memory) 

**Error Cases:**
* `--sub-path` value provided doesn't match a directory under what's provided for --local-path
* `--sub-path` and `local-path` point to same directory
* `--sub-path` value provided without  `--git*` or `--local-path`
* `--sub-path` points to an individual file which isn't a JAR/ZIP (discuss - is a single file sufficient to run a workload?)

**Autocomplete:**
* `--local-path` (to a directory on the filesystem) -- auto-complete sub-directories under the --local-path directory
* `--local-path` (to a JAR/Zip) -- auto-complete to directory structure (discuss implementation complexity/cost)
* `--git*` -- auto-complete sub-directory (discuss implementation complexity/cost)

* (sub-section problem) Pull code down and deploy the subsection locally
* (rebuilding on changes problem) Could use `--git-sha` to manually specify which commit to build from
* sub-path was legitimate in original workload creation, then in subsequent update/apply, refers to a `--local-path` directory which doesn't include the directory originally set for sub-path -- what happens?
* sub-path no longer exists in gitHub after a commit has been made -- what happens?

### Next Steps:
[ ] Additional feedback on how common it is to have multiple buildable workloads in a single repo to better understand the usecase(s).
[ ] Investigate feasibility of the underlying components (build service, supply chain) and how they might support the described behavior

### Additional context
Issue originally submitted by @danfein

shashwathi:
This PR adds support for subPath field for sources

Fixes #35


docs/command-reference/tanzu_apps_workload_apply.md
```diff
@@ -49,6 +49,7 @@ tanzu apps workload apply --file workload.yaml
       --request-memory bytes           the minimum amount of memory required, in bytes (500Mi = 500MiB = 500 * 1024 * 1024)
       --service-ref object reference   object reference for a service to bind to the workload "database=rabbitmq.com/v1beta1:RabbitmqCluster:my-broker", or "database-" to delete. This flag may be specified multiple times.
       --source-image image             destination image repository where source code is staged before being built
+      --subpath image                  subpath for git-* or image repository
```
shashwathi, line 52: @danfein : Suggestion on subpath help text would be great

pkg/commands/workload.go
```diff
@@ -155,6 +157,9 @@ func (opts *WorkloadOptions) Validate(ctx context.Context) validation.FieldError
 	if opts.SourceImage != "" {
 		source = append(source, flags.SourceImageFlagName)
 	}
+	if opts.SubPath != "" && len(source) != 1 {
+		errs = errs.Also(validation.ErrDisallowedFields(flags.SubPathFlagName, "subPath flag cannot be used without local-path/git-* sources"))
+	}
```
scothis, line 162: This check needs to be lazy. It should be valid to specify `--subpath` for an update when there is already source defined, or for a create with a file where the workload already has source defined.

pkg/commands/workload.go
```diff
@@ -229,6 +234,10 @@ func (opts *WorkloadOptions) ApplyOptionsToWorkload(ctx context.Context, workloa
 		workload.Spec.MergeSourceImage(opts.SourceImage)
 	}
 
+	if opts.SubPath != "" {
```
scothis, line 237: it should be possible to unset an existing subpath.

pkg/commands/workload.go
```diff
@@ -323,9 +332,13 @@ func (opts *WorkloadOptions) PublishLocalSource(ctx context.Context, c *cli.Conf
 	if !okToPush {
 		return okToPush, nil
 	}
+	localPathDir := opts.LocalPath
+	if opts.SubPath != "" {
+		localPathDir = filepath.Join(localPathDir, opts.SubPath)
+	}
```
scothis, line 338: this will end up applying the subpath twice, once when uploading the source, and again when building it the supply chain.

pkg/flags/flags.go
```diff
@@ -55,6 +55,7 @@ const (
 	ServiceRefFlagName    = "--service-ref"
 	SinceFlagName         = "--since"
 	SourceImageFlagName   = "--source-image"
+	SubPathFlagName       = "--subpath"
```
scothis, line 58: Sub path is two words

```suggestion
	SubPathFlagName       = "--sub-path"
```

pkg/commands/workload.go
```diff
@@ -476,6 +489,7 @@ func (opts *WorkloadOptions) DefineFlags(ctx context.Context, c *cli.Config, cmd
 	cmd.Flags().StringVar(&opts.GitCommit, cli.StripDash(flags.GitCommitFlagName), "", "commit `SHA` within the git repo to checkout")
 	cmd.Flags().StringVar(&opts.GitTag, cli.StripDash(flags.GitTagFlagName), "", "`tag` within the git repo to checkout")
 	cmd.Flags().StringVar(&opts.SourceImage, cli.StripDash(flags.SourceImageFlagName), "", "destination `image` repository where source code is staged before being built")
+	cmd.Flags().StringVar(&opts.SubPath, cli.StripDash(flags.SubPathFlagName), "", "subpath for git-* or `image` repository")
```
scothis, line 492:
```suggestion
	cmd.Flags().StringVar(&opts.SubPath, cli.StripDash(flags.SubPathFlagName), "", "`path` within the referenced source, typically a directory")
```

pkg/commands/workload.go
shashwathi: hmm very true.

pkg/commands/workload.go
rashedkvm: This change is probably unnecessary then—the supply chain to build the image based on the sub-path of the source directory.

pkg/commands/workload.go
rashedkvm: +1

pkg/commands/workload.go
```diff
@@ -323,9 +332,13 @@ func (opts *WorkloadOptions) PublishLocalSource(ctx context.Context, c *cli.Conf
 	if !okToPush {
 		return okToPush, nil
 	}
+	localPathDir := opts.LocalPath
```
rashedkvm, line 335: This change is probably unnecessary then—the supply chain to build the image based on the sub-path of the source directory.

rashedkvm (changes requested):

pkg/commands/workload.go
```diff
@@ -229,6 +230,8 @@ func (opts *WorkloadOptions) ApplyOptionsToWorkload(ctx context.Context, workloa
 		workload.Spec.MergeSourceImage(opts.SourceImage)
 	}
 
+	workload.Spec.MergeSubPath(opts.SubPath)
```
scothis, line 233: Only update the sub path when a user explicitly set the flag. Without this check, all invocations that don't set the `--sub-path` flag will inadvertently unset the existing value because the default is `""`

```suggestion
	if cli.CommandFromContext(ctx).Flags().Changed(cli.StripDash(flags.SubPathFlagName)) {
		workload.Spec.MergeSubPath(opts.SubPath)
	}
```

warango4 (approved):
LGTM

shashwathi:
@rashedkvm @scothis : Ready for another review

rashedkvm (approved):
Looks good.

Issue #21, opened 2022-03-17
Add source information to workload get output


heyjcollins:
### Description of problem

There have been cases in tap-assist where folks lose track of which workload is from which source, or where they use the wrong repo/branch and have a hard time troubleshooting the resulting unexpected behavior.

### Proposed solution

As an application developer, when I inspect a workload via workload get I would like to know what source code / image is currently used by the workload so that I can verify that the latest version has been picked up

As an application operator, when I inspect a workload via workload get I would like to know what source code / image is currently used by the workload so that I can trace back a given workload when troubleshooting or auditing

```gherkin
Given a workload has been deployed to my target cluster
When I inspect the workload via workload get
Then I can see what source code / image is currently used by the workload so that I can trace back a given workload when troubleshooting or auditing
```

### Describe alternatives you've considered
`tanzu apps workload get NAME --export` will return that that information, but as yaml or json - but those formats are less human friendly than providing name/value pairings.

### UX
If code is sourced from a git repo
```
$tanzu apps workload get tanzu-java-web-app -n tap-install
Name:   tanzu-java-web-app: Ready
---
lastTransitionTime:  "2022-01-05T06:57:18Z"
message:             ""
reason:              Ready [waiting for value at path]
status:              "True"
type:                Ready

Source:   https://github.com/spring-projects/spring-petclinic
Branch:   main
Tag   :   v1.1.0
Commit:   g5na5c
```

If code is sourced from local code
```
Source:        Local source code
Source-image:  Source:gcr.io/pivotal-knative/jc-apps:latest@sha256:3ba565457e5b5c64f5956bcb7ef9080fb3d4083f0a7c5826b48832507623c177
Updated:       23min 
```

If code is sourced from a prebuild image
```
Source:    Image
Image:     gcr.io/pivotal-knative/jc-apps:latest@sha256:3ba565457e5b5c64f5956bcb7ef9080fb3d4083f0a7c5826b48832507623c177 
Updated:   23min 
```

`Updated` — If possible to know, how long has it been since the last update from the source. The intent is to help folks know if their latest changes have been picked up. Talking to Wendy it sounds like this may not be easily knowable. 

### Additional context
Issue originally reported by @danfein

heyjcollins:
comment originally submitted by @scothis:

@danfein please close this issue if the support as implement is sufficient for your ask.

heyjcollins:
comment originally submitted by @danfein:

[...]

Issue #10, opened 2022-03-11
Display sub-path in workload get output


danfein:
### Description of problem

Workload get output includes information about the source code, but does not include a sub-path if specified. This could lead to confusion in cases where a given repository contains more than one microservice/application.

### Proposed solution

```gherkin
Given I create an workload that specifies a sub-path in the workload.yaml or via cli flag
When I `get` details about that workload
Then I should see the sub-path in the section that includes data about the source code.
```

### Example

If code is in a repo
```
$tanzu apps workload get NAME

Name:   tanzu-java-web-app: Ready
---
lastTransitionTime:  "2022-01-05T06:57:18Z"
message:             ""
reason:              Ready [waiting for value at path]
status:              "True"
type:                Ready

source:     https://github.com/odedia/todos-pcf
sub-path:   todo-ui
branch:     main
```
If code is in an image

```
source:    Image
image:     gcr.io/pivotal-knative/jc-apps:latest@sha256:3ba565457e5b5c64f5956bcb7ef9080fb3d4083f0a7c5826b48832507623c177 
sub-path:  todo-ui
```

### Additional context
Dependent on prior story to include the source information section to the workload get output.

PR #66, opened 2022-04-07
Updating Workload fields

warango4:
Updating Workload according to Cartographer's latest version. 

Updated unit testing to match Cartographer's workload fields. 

In order to support what's wanted on issues #10 and #21, the current version of Workload used in Apps Plugin needs to be upgraded, so Source information can be shown at runtime.


pkg/apis/cartographer/v1alpha1/workload.go
```diff
@@ -82,9 +82,14 @@ type WorkloadBuild struct {
 }
 
 type WorkloadStatus struct {
-	ObservedGeneration int64              `json:"observedGeneration,omitempty"`
-	Conditions         []metav1.Condition `json:"conditions,omitempty"`
-	SupplyChainRef     ObjectReference    `json:"supplyChainRef,omitempty"`
+	OwnerStatus `json:",inline"`
```
scothis, line 85: What you have is fine, but something to keep in mind is that you can also manipulate the structs a little bit to make things easier to consume, as long as the marshaled yaml is the same.

For example, instead of depending on the inlined type, you could expand it here. That way you don't need to add the extra level of nesting for OwnerStatus in all of the tests.

```suggestion
	// ObservedGeneration refers to the metadata.Generation of the spec that resulted in
 	// the current `status`.
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`

 	// Conditions describing this resource's reconcile state. The top level condition is
 	// of type `Ready`, and follows these Kubernetes conventions:
 	// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
```

Neither approach is right or wrong, but there is a choice.

pkg/apis/cartographer/v1alpha1/workload.go
```diff
@@ -82,9 +82,14 @@ type WorkloadBuild struct {
 }
 
 type WorkloadStatus struct {
-	ObservedGeneration int64              `json:"observedGeneration,omitempty"`
```
rashedkvm, line 85: After removing these - will the apps plugin work with an older version of Cart Workload API? Is this change backward compatible?

scothis: it was [moved](https://github.com/vmware-tanzu/apps-cli-plugin/pull/66/files#diff-7d0ecc1b67d2a8c12986aba3606086e9a310664b69e606aa11bccfb1170e1f35R23-R32), not removed

pkg/apis/cartographer/v1alpha1/workload.go
rashedkvm: +1

pkg/apis/cartographer/v1alpha1/workload_test.go
```diff
@@ -1489,11 +1491,13 @@ func TestWorkloadReadyConditionFunc(t *testing.T) {
 				Name:      workloadName,
 			},
 			Status: WorkloadStatus{
-				Conditions: []metav1.Condition{
-					{
-						Type:    WorkloadConditionReady,
-						Status:  metav1.ConditionFalse,
-						Message: "something went wrong",
+				OwnerStatus: OwnerStatus{
```
rashedkvm, line 1494: Treating `OwnerStatus` as type in test is not correct

shashwathi (approved):

PR #73, opened 2022-04-12
Update output for list when there are no workloads

### LINKED ISSUE #12: add `--output json` to `workloads list` command

shashwathi:
Pull request

Previously empty list would generate `null` as output, this is updated to print empty array for both json and yaml.

Fixes #12


rashedkvm (approved):
LGTM

PR #76, opened 2022-04-19
Support providing serviceAccountName with create/update/apply commands

### LINKED ISSUE #64: workload create/update/apply should support providing the spec.ServiceAccountName via workload.yaml
### Description of problem
In a majority of cases an app dev or app ops person will create/update workloads using the default supply chains available on their workload cluster, but there are edge-cases whereby a special supply chain need be invoked for which an additional, optional, workload property (`spec.serviceAccountName`) must be set to provide the supply chain with the credentials required to for the supply chain to stamp out the resources as expected.

The `apps` plugin doesn't support setting this property currently.
The only means to do so is by updating `workload.yaml` with the spec.serviceAccountName property/value out-of-band via `kubectl`. This is a sub-optimal experience. It would be ideal if the user wishing to set this workload property didn't have to switch between CLIs.

The need for setting this property isn't likely to be pervasive and persistent and the current thinking is it may not qualify for support via command flag, but it's important enough to warrant supporting the property through passing in the `workload.yaml` itself.

### Proposed solution

```gherkin
Given I want to set the `spec.serviceAccountName` value for my workload
When I execute `workload create/update/apply -f workload.yaml` and the spec.serviceAccountName property/value is set in my workload.yaml 
Then the workload object that's submitted by the apps plugin will included that property/value in the workload object it submits to the cluster
```

```gherkin
Given I want to set the `spec.serviceAccountName` value for my workload
When I search for instructions on how to do so via the apps plugin in published documentation 
Then I'm able to discover said instructions for updating/submitting the property/value via my workload.yaml
```

### Example
**workload.yaml snippet**
```
---
apiVersion: carto.run/v1alpha1
kind: Workload
metadata: {}
spec:
...
  serviceAccountName: <string>
...
```

provide documentation instructing the user to use `kubectl`.

warango4:
Pull request

Add support to se serviceAccountName property in a workload yaml file. 
It can be added while created or added after creation in yaml file and using update/apply commands for the workload to change

Fixes #64 

- Unit testing inside source code.
- Created a workload with an example yaml file with the property set. 
- Created a workload with an example yaml file without the property, then changed it to add it as part of the `spec`, used both, `update` and `apply` commands to check if the change was taken.


warango4:
Create a workload with `serviceAccountName` in the `spec` section

image

warango4:
Delete `serviceAccountName` from `spec` section

image

warango4:
Change current `serviceAccountName` using `apply`

image

odinnordico (approved):
LGTM

rashedkvm (commented):
Please see comment [here](https://github.com/vmware-tanzu/apps-cli-plugin/issues/64#issuecomment-1103171519). Add test shows removing `spec.serviceAccountName` as required by [comment](https://github.com/vmware-tanzu/apps-cli-plugin/issues/64#issuecomment-1103171519)

rashedkvm (commented):
for workload create/update/apply In the supplied yaml has `serviceAccountName=""`, remove the `serviceAccountName` field from workload.

warango4:
Deleting `serviceAccountName` by setting it to `serviceAccountName="" `

image

rashedkvm (approved):

heyjcollins:
### Acceptance
Looks good overall however...

if I execute any update that doesn't include `spec.serviceAccountName` key/value explicitly in the incantation, `spec.serviceAccountName` the plugin removes it from the workload object.
The plugin should only remove it when `spec.serviceAccountName: ""` is provided.

PR #77, opened 2022-04-19
Adding yml to accepted output formats

### LINKED ISSUE #72: `workload get -o` and `workload list -o` accepts `yaml` and/or `yml`
### Description of problem
Workload get and workload list commands accept`yaml` as an `--output` value today.
Because `yml` is a valid extension for the same file-type, we should accept `yml` and/or `yaml`

### Proposed solution

```gherkin
Given I have a workload named foo
When I run `tanzu apps workload get foo -o yml`
Then the output from the command is exactly the same as if I ran `tanzu apps workload get foo -o yaml`
```

```gherkin
Given I have created workloads on my target cluster
When I run `tanzu apps workload list -o yml`
Then the output from the command is exactly the same as if I ran `tanzu apps workload list -o yaml`
```

warango4:
Pull request

Add support for yml extension when user wants to output a resource through workload get or workload list commands.

Output will be same if user wants to show it with yaml format.

Usage example:
`tanzu apps workload get my-workload -oyml`
`tanzu apps workload list -oyml`

Fixes #72 

- Unit testing in source code
- Used commands in local cluster

### DIFF
--- a/docs/command-reference/tanzu_apps_workload_get.md
+++ b/docs/command-reference/tanzu_apps_workload_get.md
@@ -22,7 +22,7 @@ tanzu apps workload get my-workload
       --export           export workload in yaml format
   -h, --help             help for get
   -n, --namespace name   kubernetes namespace (defaulted from kube config)
-  -o, --output string    output the Workload formatted. Supported formats: "json", "yaml"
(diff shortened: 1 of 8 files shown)


odinnordico (approved):
LGTM

rashedkvm (approved):

PR #86, opened 2022-04-28
Removing namespace from service reference is not ignored

### LINKED ISSUE #17: workload update/apply - the removal of namespace from previously created --service-ref is ignored with incorrect message

- [X] I reviewed open and closed Github issues that may be related to my problem.
- [X] I am reporting a bug that others will be able to reproduce.

### Describe the bug
When user removes a namespace from a previously created --service-ref entry which originally included namespace, the apps plugin doesn't think there's anything to change so the update is skipped.

### Expected behavior
removal of namespace should trigger an update to the --service-ref claim

### Steps to Reproduce
1. add the following --service-ref to an existing workload:
  * `tanzu apps workload apply [my-app] --service-ref my-db=rabbitmq.com/v1beta1:RabbitmqCluster:my-namespace:my-broker`
1. update the `my-db` service-ref (remove `:my-namespace`):
  * `tanzu apps workload apply [my-app] --service-ref my-db=rabbitmq.com/v1beta1:RabbitmqCluster:my-broker`

The result of the update should be that the my-namespace portion of the service-ref should be removed (after the diff is displayed and user-prompt accepted)

apps version v0.4.0+

TAP v0.5.0 on GKE

shashwathi:
Fix the bug when namespace is deleted from service reference and it was ignored in annotations. With this fix, annotations will be updated correctly.

Fixes #17 

Use the test cases in github issue

### DIFF
--- a/pkg/commands/workload.go
+++ b/pkg/commands/workload.go
@@ -32,7 +32,6 @@ import (
 
 	"github.com/vmware-tanzu/apps-cli-plugin/pkg/apis"
 	cartov1alpha1 "github.com/vmware-tanzu/apps-cli-plugin/pkg/apis/cartographer/v1alpha1"
-	servicev1alpha1 "github.com/vmware-tanzu/apps-cli-plugin/pkg/apis/services/v1alpha1"
 	cli "github.com/vmware-tanzu/apps-cli-plugin/pkg/cli-runtime"
 	"github.com/vmware-tanzu/apps-cli-plugin/pkg/cli-runtime/parsers"
 	"github.com/vmware-tanzu/apps-cli-plugin/pkg/cli-runtime/validation"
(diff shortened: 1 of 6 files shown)


pkg/commands/workload.go
```diff
@@ -256,28 +255,24 @@ func (opts *WorkloadOptions) ApplyOptionsToWorkload(ctx context.Context, workloa
 		}
 	}
 
-	sc := servicev1alpha1.NewServiceClaimWorkloadConfig()
 	for _, ref := range opts.ServiceRefs {
 		parts := parsers.DeletableKeyValue(ref)
+		deleteKey := parts[0]
```
rashedkvm, line 260: **nit**: This key is used for Merge and Delete
```suggestion
		serviceRefKey := parts[0]
```

PR #87, opened 2022-04-28
Do not delete ServiceAccountName when updating other fields

### LINKED ISSUE #78: ServiceAccountName is removed for apply/update workloads commands unless included in a file

### Describe the bug

For a workload with a `.spec.serviceAccountName` set, when executing an apply/update command the service account is unset.

```
tanzu apps workload update petclinic --label hello=world
Update workload:
...
  3,  3   |kind: Workload
  4,  4   |metadata:
  5,  5   |  labels:
  6,  6   |    apps.tanzu.vmware.com/workload-type: web
      7 + |    hello: world
  7,  8   |  name: petclinic
  8,  9   |  namespace: default
  9, 10   |spec:
 10, 11   |  env:
 11, 12   |  - name: SPRING_PROFILES_ACTIVE
 12, 13   |    value: mysql
 13     - |  serviceAccountName: default
 14, 14   |  serviceClaims:
 15, 15   |  - name: db
 16, 16   |    ref:
 17, 17   |      apiVersion: v1
...

? Really update the workload "petclinic"? No
```

### Expected behavior

One should be able to execute an update/apply command against that workload to modify another value without the service account being unset.

### Steps to Reproduce

1. create a workload with a service account name2. 
2. run any `tanzu apps workload apply/update` command that would change the workload and doesn't also include the `--file` flag. 

Custom build from main HEAD fc7b7180fe43523645976fcb0fb79aaf8b0f95d6

warango4:
Pull request

ServiceAccountName should be added, updated or deleted through a yaml file. If workload is being changed through flags
and there is no change in yaml file related to ServiceAccountName, it shouldn't be removed.
This commit fixes the introduced bug.

Fixes #78 

- Created and updated different workloads in local cluster.
- Added unit tests

Check #64 and #76 for more context.


warango4:
image

warango4:
image

warango4:
image

shashwathi (changes requested):
Existing changes look good. Can you add the same logic for create as well? if user can create workload via `apply` command, SA will not be deleted but it will be for `create` command so I would recommend adding support for update, create and apply sub-commands.

warango4:
Create workload with `serviceAccountName`

image

warango4:
Create workload with `serviceAccountName` and check fields passed through flags do not override/delete this field

image

scothis:
Another option is to make the ServiceAccountName field type a string pointer (`*string`). This way unchanged (`nil`) can be distinguished from unset (`""`). It's a bit more annoying to work with, but can help avoid the ambiguity.

shashwathi (approved):

warango4:
I would keep this change as is and do a follow up PR with changes suggested by Scott.

rashedkvm (approved):

heyjcollins:
validated serviceAccountName isn't omitted on workload create, apply

PR #110, opened 2022-05-09
subpath field is not removed when submitting workload from file

### LINKED ISSUE #88: --sub-path is removed when submitting the exact workload.yaml which includes it already

- [X] I reviewed open and closed Github issues that may be related to my problem.
- [X] I am reporting a bug that others will be able to reproduce.

### Describe the bug
applying or updating the exact workload.yaml that was previously exported which includes  a `spec.source.subPath` property/value --- causes the plugin to remove `subPath` from the workload

### Expected behavior
subPath shouldn't be removed

### Steps to Reproduce
create a workload:
```
tanzu apps workload create pc --git-repo https://github.com/spring-projects/spring-petclinic --gi
t-branch main --sub-path foo --service-ref database=rabbitmq.com/v1beta1:RabbitmqCluster:my-broker --limit-cpu 500 --annotation anba
r=bar --param foo=bar --label lfoo=lbar
```

export the workload:
```
tanzu apps workload get pc --export > pc.yaml
```

apply the workload.yaml that was just created in previous step (without changing it at all)
```
tanzu apps workload apply -f pc.yaml
```

You should see the removal of `subPath` in the resulting diff.

```sh
``` 
version: v0.11.2
buildDate: 2022-03-17
sha: 9f16f375
v0.0.0-dev ---- the apps plugin was built locally from this commit: 6ae1969b2a495eb29be08d63bf6c0c72cce83981

tap installed on GKE
client - Mac

shashwathi:
Pull request
subpath field is not ignored when submitting workload from file(both while creating & updating)

Fixes #88 

- create workload with subPath
- export the workload into a file
- apply the workload again, subPath should not be deleted

### DIFF
--- a/pkg/commands/testdata/workload-subPath.yaml
+++ b/pkg/commands/testdata/workload-subPath.yaml
@@ -1,0 +1,25 @@
+# Copyright 2021 VMware, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: carto.run/v1alpha1
+kind: Workload
+metadata:
+  name: service
+spec:
+  source:
+    subPath: ./app
+    git:
+      url: https://github.com/spring-projects/spring-petclinic.git
+      ref:
+        branch: main
--- a/pkg/commands/workload_apply_test.go
+++ b/pkg/commands/workload_apply_test.go
@@ -173,6 +173,94 @@ Created workload "my-workload"
 `,
 		},
 		{
+			Name: "Update git source with subPath from file",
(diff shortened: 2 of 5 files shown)


pkg/commands/workload_apply_test.go
```diff
@@ -169,6 +169,94 @@ Create workload:
...
+ 10, 10   |      ref:
+ 11, 11   |        branch: main
+ 12, 12   |      url: https://github.com/spring-projects/spring-petclinic.git
+     13 + |    subPath: ./app
+
+Updated workload "my-workload"
+`,
+		},
+		{
+			Name: "Update git source with subPath from file",
```
warango4, line 223: I think this one should be called `Create git source with subPath from file`

warango4, line 223: I think this one should be called `Create git source with subPath from file`

warango4 (approved):

atmandhol:
Minor nit, when we merge this, we should change the commit message from `[fix]` to `fix:` to follow the standard.

pkg/commands/testdata/workload-subPath.yaml
```diff
@@ -0,0 +1,25 @@
+# Copyright 2021 VMware, Inc.
```
rashedkvm, line 1: Issue: should we update the template https://github.com/vmware-tanzu/apps-cli-plugin/blob/baf4a47308c2b1615935a97010318d41864ac2a3/hack/boilerplate.go.txt
```
# Copyright 2021 VMware, Inc.
```

PR #111, opened 2022-05-10
Dependency upgrades

shashwathi:
- Update dies from 0.3 to 0.4
- Update k8s client, api, apimachinery frm 0.23.6 to 0.24.0

Note: 

k8s.io/apiextensions-apiserver version is not upgraded because tanzu-framework is dependent on 0.23 so that is still retained. Rest of the k8s lib are upgraded


rashedkvm (approved):
Thank you @shashwathi for cleaning it up.

PR #121, opened 2022-05-17
Add support for jar/zip format for local-path flag

### LINKED ISSUE #29: --local-path supports jar/zip
### Description of problem

There are processes/pipelines being utilized today which generate a jar/zip file artifact.

These pipelines will often pull in additional assets which aren't included in the application code itself, but which are necessary for the application function.
A subsequent next step is for the jar/zip of the app to be pushed to and run on a target environment.
Popular platforms such as Cloud Foundry support this workflow today (via the cf CLI). 

The apps plugin for the Tanzu CLI _does not_ support creating/updating/applying a workload from a local jar or zip directly.
The file must be unzipped beforehand. 
Requiring the jar be unzipped is not an expected step, it creates toil without value, and decreases efficiency.

### Proposed solution

```gherkin
Given I've got a JAR (or zipped) file containing all assets required to create a workload and I'm targeting a TAP-enabled workload cluster
When I run `tanzu apps workload create myapp -f path/to/my/myapp.jar...`
Then my workload create command exits 0 and I can validate the workload has run through the supply chain by viewing the workload details via `tanzu apps workload get...`
```

@danfein to provide UX here

### Additional context
* cc: @matthewmcnew @jigsheth57 may have additional context/thoughts on this issue
* from @scothis: 
  * It's worth mirroring the behavior of pack. https://github.com/buildpacks/pack/blob/3fc8e2220b584da76cd0227bbc211aae787e0e57/pkg/client/build.go#L596-L605

shashwathi:

Add support for jar/zip format for local-path flag

Fixes #29 

Download simple java jar file and build zip file(using `zip` cli).


pkg/commands/workload.go
```diff
@@ -324,8 +324,23 @@ func (opts *WorkloadOptions) PublishLocalSource(ctx context.Context, c *cli.Conf
...
+	} else if source.IsZip(opts.LocalPath) {
+		zipContentsDir, err := source.HandleZip(opts.LocalPath)
+		if err != nil {
+			c.Errorf("Failed to extract file contents from %q. \n", opts.LocalPath)
+			return false, err
+		}
+		defer os.RemoveAll(zipContentsDir)
+		contentDir = zipContentsDir
+	} else {
+		return false, fmt.Errorf("unsupported file format %q", opts.LocalPath)
```
shashwathi, line 339: cc @danfein : this is the error if user provides a file which is not a dir and not a valid zip file. If you want something else, please add the msg here

pkg/source/archive.go
```diff
@@ -0,0 +1,138 @@
...
+	"archive/zip"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+)
+
+func HandleZip(fileName string) (string, error) {
+	tmpDir, err := ioutil.TempDir("", "")
```
rashedkvm, line 29: issue: When is the temp directory removed?
```
	tmpDir, err := ioutil.TempDir("", "")
```

pkg/source/archive.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		if err := srcFile.Close(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func IsZip(fileName string) bool {
```
rashedkvm, line 99: Question: `IsZip` testing both "jar" and "zip" file types?

pkg/commands/workload.go
```diff
@@ -324,8 +324,23 @@ func (opts *WorkloadOptions) PublishLocalSource(ctx context.Context, c *cli.Conf
...
+	var contentDir string
+	if source.IsDir(opts.LocalPath) {
+		contentDir = opts.LocalPath
+	} else if source.IsZip(opts.LocalPath) {
+		zipContentsDir, err := source.HandleZip(opts.LocalPath)
+		if err != nil {
+			c.Errorf("Failed to extract file contents from %q. \n", opts.LocalPath)
+			return false, err
+		}
+		defer os.RemoveAll(zipContentsDir)
```
warango4, line 336: > issue: Shouldn't the temp directory be cleaned up at some point (removed)?
@rashedkvm I think that's being done here.

warango4 (approved):

pkg/source/archive.go
shashwathi: So we need the temp dir for imgpkg to package up the contents. So in the called function, tmp dir is cleanedup.

scothis: it would be better to accept the directory as an argument, and then the cleanup is clearly the callers problem

rashedkvm (approved):

PR #142, opened 2022-05-27
Add new command cluster-supply-chain get

### LINKED ISSUE #134: New - Cluster-supply-chain get command
### Description of problem

Folks can list the supply chains available on a cluster with `tanzu apps cluster-supply-chain get`, but there is not currently a good way of investigating the details of the supply chains within the tanzu CLI.

In particular, there is not a good way for application developers to know what the selectors a given supply chain uses are.

### Proposed solution

```gherkin
Given I need information about a supply chain 
When I request information via the CLI (proposed `cluster-supply-chain get`)
Then the CLI should return details about the requested supply chain.
```
Proposed - A new command to get information about supply chains.

There are a number of things that would be helpful to surface, for the first draft of this feature the supply chain selector is a useful bit of information that would be straightforward to retrieve / present, and would help folks better understand how a given chain chooses workloads to pickup. 

### Example

Supply chain with no selectors (shouldn't happen, but you never know)

```
❯ tanzu apps cluster-supply-chain get source-to-url
---
# source-to-url               
---
Supply Chain Selectors                     
No supply chain selectors found
```

Supply chain with one selector

```bash
❯ tanzu apps cluster-supply-chain get source-to-url
---
# source-to-url               
---
Supply Chain Selectors                     
TYPE         KEY             OPERATOR   VALUE
label        workload-type              web

```

Supply chain with multiple selectors 

```
❯ tanzu apps cluster-supply-chain get basic-image-to-url
---
# basic-image-to-url               
---
Supply Chain Selectors                     
TYPE         KEY             OPERATOR   VALUE
label        workload-type              web
fields       spec.image      exists
expressions  group           in         public facing
expressions  group           in         internal

```
Note, in the case of an array, each value will show up on a new line rather than combining them in a single line item (ie expressions group in {"public facing", "internal"} would be rendered as shown above)

docs + kubeconfig / k9s to interrogate supply chains

### Additional context
@atmandhol and I looked over the template selectors, and would consider including them in a future release. Because of the quick turn around on the first release of this feature, and the more complicated challenge of clearly displaying the template selectors, AND because we have not seen supply chains using them yet, we chose to defer that information for this release.

shashwathi:
Add new command cluster-supply-chain get

Fixes #134


shashwathi:
Need to investigate more about why selectors not being recognized with API definition.

PR #145, opened 2022-05-27
feat: added suport for exclude files on local path

### LINKED ISSUE #33: create/update/apply supports --exclude
### Description of problem
raw feedback below --- requires polishing.
File excludes; as more systems around us adopt the "as code" approaches, developers will increasingly have files in their projects that aren't anything do with actually running code; i.e. they dont end up in the running container this impacts the usability of --local-path as it sends the source more often than it needs to. This situation already exists actually as dev's often store project specific IDE configuration in their projects. Also, bare in mind the new backstage component.yaml/system.yaml. All nothing to do with running the app. but devs often change them and when they next run apply with -local-path=. publishes the source code image. But in reality the app source code hasn't actually changed.

As more systems around us adopt the "as code" approaches, application developers will increasingly have files in their projects that have nothing to do with actually running code (those files don't end up in the running container).

This impacts the usability of `--local-path` as it:
* consumes resources unnecessarily (uploads code that isn't required)
* triggers automatic deployments (when `--live-update` is enabled) whenever any of those unnecessary files change. 

### Proposed solution

```gherkin
Given I have an app project that includes extraneous `.foo` files 
When I create a workload with my project and enable `--live-update`
Then my `*.foo` files are not uploaded on initial creation and when I update any `.foo` file in my project, a live-update is not triggered.
```
The UX needs to be designed with @danfein (needs-design label applied).

Decision on solution hasn't been made...
* use an --exclude flag on a per-command basis
* use a persistent .ignore file

#### Prior art for review:
- https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files
- https://www.perforce.com/manuals/p4vs/Content/P4VS/managing.excluding.html
- https://tortoisesvn.net/docs/release/TortoiseSVN_en/tsvn-dug-ignore.html 
- https://marketplace.visualstudio.com/items?itemName=AdamWulkiewicz.ExcludeFromBuild
- https://www.jetbrains.com/help/idea/compiler-excludes.html

odinnordico:
Pull request

Add support to exclude files when creating/updating a workload from `--local-source` using the file `.tanzuignore`.

Fixes #33 

Testing on local box against Tap 1.0.2 cluster with local java project

image

image

`tanzu apps wld apply tanzu-java-web-app-2-6-0 --local-path . --source-image dalfonso901/tanzu-java-web-app-2-6-local --label autoscaling.knative.dev/min-scale=1 --type web`

image

image

image

image

There is a missing integration test to verify the final image comes without excluded files


anibmurthy (approved):
lgtm

pkg/commands/workload.go
```diff
@@ -390,6 +402,40 @@ func (opts *WorkloadOptions) checkToPublishLocalSource(taggedImage string, c *cl
...
+	exclude := []string{}
+	if opts.ExcludePathFile != "" {
+		p := filepath.Join(opts.LocalPath, opts.ExcludePathFile)
+		if _, err := os.Stat(p); errors.Is(err, os.ErrNotExist) {
+			return exclude
+		}
+
+		f, err := os.Open(p)
+		if err != nil {
+			c.Infof("Unable to read %s file.\n", opts.ExcludePathFile)
```
shashwathi, line 415: is it possible to add a test for this? (have a file with settings that it cant open with chmod set for root user possibly)

pkg/commands/workload.go
```diff
@@ -390,6 +402,40 @@ func (opts *WorkloadOptions) checkToPublishLocalSource(taggedImage string, c *cl
...
+		for {
+			l, _, err := r.ReadLine()
+			if err == io.EOF {
+				break
+			}
+			p := strings.TrimSpace(string(l))
+			if len(p) == 0 || strings.HasPrefix(p, "#") {
+				continue
+			}
+			if strings.HasSuffix(p, string(os.PathSeparator)) {
```
shashwathi, line 429: why is PathSeparator stripped out? 
- i think its because imgpkg is doing exact match on the name of dir, i want to confirm my understanding here

pkg/commands/workload_test.go
```diff
@@ -1610,6 +1611,16 @@ func TestWorkloadOptionsPublishLocalSource(t *testing.T) {
 		shouldError    bool
 		expectedOutput string
 	}{{
+		name:     "local source with excluded files",
```
shashwathi, line 1614: can you add a test or update this test with tanzuignore file 
- that includes a dir path that exists
- a nested file in the dir thats excluded

shashwathi (commented):
Overall looks good. I would add some more unit tests to make sure we have coverage for diff file paths in tanzuignore file

pkg/commands/workload.go
odinnordico: yes, that is correct, folders suffixed with `PathSeparator` are not being excluded

pkg/commands/workload_test.go
odinnordico: Test updated

shashwathi (approved):

pkg/commands/workload_update.go
```diff
@@ -184,7 +184,11 @@ func (opts *WorkloadUpdateOptions) IsDryRun() bool {
 }
 
 func NewWorkloadUpdateCommand(ctx context.Context, c *cli.Config) *cobra.Command {
-	opts := &WorkloadUpdateOptions{}
+	opts := &WorkloadUpdateOptions{
+		WorkloadOptions: WorkloadOptions{
```
shashwathi, line 188: I would recommend adding LoadDefaults func on `WorkloadOptions`. In that func, you can set ExcludeFilePath to the default file. Declare the default path of tanzuignore as const in workload.go file. In this pattern, you can remove TanzuIfnoreFile from cli config completely.  WDYT?

odinnordico: I like the idea of `LoadDefaults` function but the `cli.Config`'s `TanzuIfnoreFile` is to be used later as a global cli configuration and i see more sense in cofig besides to kubecofig file and viper config file

odinnordico: Added `LoadDefaults` *func* to `WorkloadOptions` and called on each sub-command

pkg/commands/workload_test.go
```diff
@@ -1597,6 +1597,7 @@ func TestWorkloadOptionsApplyOptionsToWorkload(t *testing.T) {
 func TestWorkloadOptionsPublishLocalSource(t *testing.T) {
 	registry, err := ggcrregistry.TLS("localhost")
 	utilruntime.Must(err)
+
```
shashwathi, line 1600: nit: remove new line

shashwathi (approved):
/lgtm

PR #156, opened 2022-06-03
Add a msg for workload sub commands when no source is present

### LINKED ISSUE #154: Print info message to inform users when source cannot be validated
### Description of problem

The current implementation of the maven source uses the params to specify the source details. 
The CLI is currently validating the source section/flags to let folks know as early as possible if a workload is incomplete.
Because the maven source is not in the source section, the CLI validation is preventing maven workloads from being created.

### Proposed solution

```gherkin
Given A workload is created/updated
When No source is specified 
Then The workload will not be blocked by validation and a notice will be printed with the diff
```

### Example

```
❯ tanzu apps workload create test-name 
Create workload:
      1 + |---
      3 + |kind: Workload
      4+ | params: <maven content>
       ...

NOTICE: no source code or image has been specified for this workload.

? Do you want to create this workload? (y/N)
```

--no-validate, to disable validation of inputs 
  disabling all validation makes the flag usable for a number of situations, but may not be desirable just to enable non-standard workload sources

--skip-source-check, to disable the validation of source inputs
  Not ideal to have a flag for such a narrow use-case, particularly if maven may move to the standard source section in future releases.

--source-yaml, accepts same input as --params-yaml, and also disables source validation
  Similar drawbacks as skipping the source check via a flag, essentially duplicates an existing flag to achieve the same outcome

parsing the params section to validate that it contains source information 
  Too brittle, and would put us on the hook for parsing future non-standard source types

shashwathi:
Add a info level msg for workload create/apply/update when no source is present

Fixes #154 

Screen Shot 2022-06-03 at 1 25 17 PM

- if workload is created with no source but subPath is provided : fails

### DIFF
--- a/pkg/apis/cartographer/v1alpha1/workload_helpers.go
+++ b/pkg/apis/cartographer/v1alpha1/workload_helpers.go
@@ -106,17 +106,30 @@ func (w *Workload) Validate() validation.FieldErrors {
 	return errs
 }
 
+func (w *Workload) IsSourceFound() bool {
+	if w.Spec.Source == nil && w.Spec.Image == "" {
+		return false
+	}
+	if w.Spec.Source != nil {
+		if w.Spec.Source.Git == nil && w.Spec.Source.Image == "" {
+			return false
+		}
+	}
+	return true
+}
+
 func (w *WorkloadSpec) Validate() validation.FieldErrors {
 	errs := validation.FieldErrors{}
-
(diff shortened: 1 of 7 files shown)


odinnordico (approved):
LGTM

warango4 (approved):
Lgtm

anibmurthy (approved):
lgtm

atmandhol (approved):

PR #163, opened 2022-06-07
Add hint message in workload create/update/apply/get

### LINKED ISSUE #150: Append "next command recommendations" to the output from `workload create/update/apply/get`
### Description of problem
App devs experiencing issues with their workload (e.g. not making it through a supply chain) get stuck while troubleshooting and don't readily/easily discover the `tail` command and as a result quickly learn to use other tools while troubleshooting.

We can address this problem incrementally, and a quick/easy way to begin doing so it to recommend certain commands that would logically be executed next in the output.

### Proposed solution

1. Append a recommendation to run `tanzu apps workload tail YOUR-APP` to the current output for `workload create/update/apply AND get`
2. Append a recommendation to run `tanzu apps workload get YOUR-APP` to the current output for `workload create/update/apply`

### Example
**tanzu apps workload create/update/apply**
```
%tanzu apps workload create/update/apply my-app --local-path ~/workspace/tanzu-java-web-app --type web --source-image gcr.io/pivotal-knative/jc-apps
? Publish source in "/Users/collinsjo/workspace/tanzu-java-web-app" to "gcr.io/pivotal-knative/jc-apps"? It may be visible to others who can pull images from that repository Yes
Publishing source in "/Users/collinsjo/workspace/tanzu-java-web-app" to "gcr.io/pivotal-knative/jc-apps"...
Published source
Create workload:
      1 + |---
      3 + |kind: Workload
      4 + |metadata:
      5 + |  labels:
      6 + |    apps.tanzu.vmware.com/workload-type: web
      7 + |  name: my-app
      8 + |  namespace: ns1
      9 + |spec:
     10 + |  source:
     11 + |    image: gcr.io/pivotal-knative/jc-apps:latest@sha256:22f23f4f143ce371fbce513601423f7965ff070572882c200a3191374f4daf56

? Do you want to create this workload? Yes
Created workload "my-app"

Tail workload logs by running "tanzu apps workload tail my-app --since 120m" (adjust "--since" value as needed)
View a snapshot of the workload state by running "tanzu apps workload get my-app"

%|
```

**tanzu apps workload get**
```
%tanzu apps workload get my-app
# my-app: Unknown
---
Source
type:    source image
image:   gcr.io/pivotal-knative/jc-apps:latest@sha256:22f23f4f143ce371fbce513601423f7965ff070572882c200a3191374f4daf56

Supply Chain
name:          source-to-url
last update:   6m32s
ready:         Unknown

RESOURCE          READY   TIME
source-provider
deliverable
image-builder
config-provider
app-config
config-writer

Issues
reason:    MissingValueAtPath
message:   waiting to read value [.status.latestImage] from resource [image.kpack.io/my-app] in namespace [ns1]

Pods
NAME                       STATUS   RESTARTS   AGE
my-app-build-1-build-pod   Failed   0          6m31s

Tail workload logs by running "tanzu apps workload tail my-app --since 120m" (adjust "--since" value as needed)

%|
```

### Additional context
cc @danfein - please review and make any UX suggestions/alterations

warango4:
Pull request

Add hint message for commands that can be used for troubleshooting or just check status of a workload.

- For create/update/apply: add hint so user can check logs and status through tail and get commands
- For get: add hint so user can check logs through tail command

Fixes #150 

- Unit testing
- Test in local cluster

Signed-off-by: Wendy Arango warango@vmware.com

### DIFF
--- a/pkg/commands/workload_apply.go
+++ b/pkg/commands/workload_apply.go
@@ -150,8 +150,14 @@ func (opts *WorkloadApplyOptions) Exec(ctx context.Context, c *cli.Config) error
 		}
 	}
 
+	if okToCreate || okToUpdate {
+		c.Printf("\n")
+		DisplayCommandNextSteps(c, workload)
+		c.Printf("\n")
+	}
+
 	anyTail := opts.Tail || opts.TailTimestamps
-	if (opts.Yes || okToCreate || okToUpdate) && (opts.Wait || anyTail) {
+	if (okToCreate || okToUpdate) && (opts.Wait || anyTail) {
 		c.Infof("Waiting for workload %q to become ready...\n", opts.Name)
 
 		workers := []wait.Worker{
--- a/pkg/commands/workload.go
+++ b/pkg/commands/workload.go
@@ -159,6 +159,16 @@ func (opts *WorkloadOptions) Validate(ctx context.Context) validation.FieldError
 	return errs
 }
 
+func DisplayCommandNextSteps(c *cli.Config, workload *cartov1alpha1.Workload) {
+	if workload.Namespace != c.Client.DefaultNamespace() {
+		c.Infof("To see logs:   \"tanzu apps workload tail %s %s %s\"\n", workload.Name, flags.NamespaceFlagName, workload.Namespace)
+		c.Infof("To get status: \"tanzu apps workload get %s %s %s\"\n", workload.Name, flags.NamespaceFlagName, workload.Namespace)
+	} else {
+		c.Infof("To see logs:   \"tanzu apps workload tail %s\"\n", workload.Name)
+		c.Infof("To get status: \"tanzu apps workload get %s\"\n", workload.Name)
+	}
+}
+
 func (opts *WorkloadOptions) LoadDefaults(c *cli.Config) {
 	opts.ExcludePathFile = c.TanzuIgnoreFile
 }
@@ -461,6 +471,8 @@ func (opts *WorkloadOptions) Update(ctx context.Context, c *cli.Config, currentW
 				return okToUpdate, nil
 			}
 		}
+	} else {
+		okToUpdate = opts.Yes
 	}
 
 	if err := c.Update(ctx, workload); err != nil {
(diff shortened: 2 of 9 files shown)


warango4:
image

warango4:
image

warango4:
image

odinnordico (approved):

pkg/commands/workload_apply.go
```diff
@@ -150,41 +150,47 @@ func (opts *WorkloadApplyOptions) Exec(ctx context.Context, c *cli.Config) error
...
-			func(ctx context.Context) error {
-				clientWithWatch, err := watch.GetWatcher(ctx, c)
-				if err != nil {
-					panic(err)
-				}
-				return wait.UntilCondition(ctx, clientWithWatch, types.NamespacedName{Name: workload.Name, Namespace: workload.Namespace}, &cartov1alpha1.WorkloadList{}, cartov1alpha1.WorkloadReadyConditionFunc)
-			},
-		}
+	if okToCreate || okToUpdate {
+		c.Printf("\n")
```
shashwathi, line 154: Can you move this logic above the `if` condition and keep rest the same? 
when I tested this locally, I noticed the test case that fails is when workload to be update has no change(noop), the helper msg is printed. That doesnt seem like a bad experience. WDYT @warango4 ?
Let me know if I am missing something

shashwathi (changes requested):

pkg/commands/workload_apply.go
warango4: From conversation with design:

>  if the person saw something that caused them to stop the update, I would guess their next step would be to change the work they were doing and not to check logs / workload get.

(...)

> it helps prevent us from spamming the users with these messeges, we should just show them in places where the user is likley to want to do the thing we are suggesting, so we don’t jsut use it to promote whatever the new feature of the day is. If we turn it into an ad, folks will tune it out.

That's why I added this after validation if user does actually want to create/update/apply their workload.

And about adding a new test case, I do not know how to, in the current way we do tests, add one where there's no usage of `--yes` flag but instead, survey is answered with `No`, so I'm not quite sure how to cover these cases. At least in `apply` and `update`, there is the `noop` test, that shows the desired behaviour, not sure how to add something similar in create.

shashwathi (approved):

rashedkvm (approved):

PR #208, opened 2022-07-05
test: added integration test suite

### LINKED ISSUE #159: Add integration e2e suite
### Description of problem

Add integration e2e suite(preferably in Golang).

- e2e test should target existing k8s cluster with cartographer CRDs installed.
- Add test for CRUD for workload
- Add test for CSC get/list commands
- Cleanup the resources as part of the test suite.

odinnordico:
# Pull request

Adds an integration test suite with some examples of how to use it

Fixes #159 

Local run of the integration test

### DIFF
--- a/testing/e2e/suite/comand_line_integration.go
+++ b/testing/e2e/suite/comand_line_integration.go
@@ -1,0 +1,215 @@
+//go:build integration
+// +build integration
+
+/*
+Copyright 2022 VMware, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package suite_test
+
+import (
+	"reflect"
+	"regexp"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	cartov1alpha1 "github.com/vmware-tanzu/apps-cli-plugin/pkg/apis/cartographer/v1alpha1"
+)
+
+var (
+	dfOpts = []cmp.Option{
+		cmp.Transformer("T1", func(in string) string {
+			s := strings.TrimSpace(in)
+			s = strings.Replace(in, "  ", " ", -1)
+			return regexp.MustCompile(`( (\d+[smhd]){1,4})[ \n$]`).ReplaceAllString(s, "0s")
+		}),
+		// ^([0-9]{1,10}[dhms]{1}){1,4}$
+	}
+	objOpts = []cmp.Option{
+		cmpopts.IgnoreFields(cartov1alpha1.Workload{},
+			"Status",
+			"GenerateName",
+			"SelfLink",
+			"UID",
+			"ResourceVersion",
+			"Generation",
+			"CreationTimestamp",
+			"DeletionTimestamp",
+			"DeletionGracePeriodSeconds",
+			"OwnerReferences",
+			"Finalizers",
+			"ZZZ_DeprecatedClusterName",
+			"ManagedFields",
+		),
+	}
+)
+
+const TestingNamespace = "apps-integration-testing"
+
+type CommandLineIntegrationTestSuite []CommandLineIntegrationTestCase
+
+type CommandLineIntegrationTestCase struct {
+	Name                      string
+	Skip                      bool
+	Focus                     bool
+	WorkloadName              string
+	Command                   CommandLine
+	ExpectedCommandLineError  bool
+	ExpectedCommandLineOutput string
+	SkipClusterValidation     bool
+	ExpectedRemoteObject      client.Object
+	IsList                    bool
+	ExpectedRemoteList        client.ObjectList
+	SleepMillisBefore         int
+	SleepMillisAfter          int
+	CheckCommandOutput        bool
+}
+
+func (ts CommandLineIntegrationTestSuite) Run(t *testing.T) {
+
+	cleanUp(t, false)
+	preapreEnvironment(t)
+	defer cleanUp(t, t.Failed())
+
+	testToRun := ts
+	focused := CommandLineIntegrationTestSuite{}
+	for _, tc := range ts {
+		if tc.Focus && !tc.Skip {
+			focused = append(focused, tc)
+		}
+	}
+	isFocused := len(focused) != 0
+	if isFocused {
+		testToRun = focused
+	}
+
+	conf, err := NewClientConfig()
+	if err != nil {
+		t.Errorf("Error creating validation config: %v", err)
+		t.FailNow()
+	}
+
+	for _, tc := range testToRun {
+		tc.Run(t, conf)
+		if t.Failed() {
+			break
+		}
+	}
+	if isFocused {
+		t.Errorf("test run focused on %d record(s), skipped %d record(s)", len(focused), len(ts)-len(focused))
+	}
+
+}
+
+func (cl CommandLineIntegrationTestCase) Run(t *testing.T, conf *ClientConfig) {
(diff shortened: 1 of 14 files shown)


testing/e2e/helpers/object.go
```diff
@@ -0,0 +1,52 @@
...
+	"reflect"
+
+	"github.com/google/go-cmp/cmp"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	cartov1alpha1 "github.com/vmware-tanzu/apps-cli-plugin/pkg/apis/cartographer/v1alpha1"
+)
+
+// TODO Improve the comparing of workloads with the autogenerated fields/dates
+func CompareObject(o1, o2 client.Object) bool {
```
shashwathi, line 32: so while comparing workload objects we care about 
- in metadata section (everything other than timestamp)
- all of spec
- ignore status
- We can use `cmp.Equal` for object metadata and pass ignoretimestamp as the option. 
-

testing/e2e/test_cases/workload_create_test.go
```diff
@@ -0,0 +1,73 @@
...
+					"--app=test-create-git-annotations-workload",
+					"--git-tag=tap-1.2",
+					"--git-repo=https://github.com/sample-accelerators/spring-petclinic",
+					namespaceFlag,
+					"--type=web",
+					"--yes",
+				)
+				return c
+			}(),
+			ExpectedRemoteObject: helpers.GetWorkloadFromFile(t, filepath.Join(expectedBasePath, "create-wld", "test-create-git-annotations-workload.yaml")),
```
shashwathi, line 48: isnt it better to build workload object here instead of reading from file? To follow the test expectations I need to open another file. for readability sake i would stick to building the workload object here,

testing/e2e/test_cases/workload_create_test.go
```diff
@@ -0,0 +1,73 @@
...
+			Command: *it.NewCommandLine(
+				"tanzu", "apps", "workload", "delete", "test-create-git-annotations-workload", namespaceFlag, "-y"),
+			ExpectedRemoteObject: emptyWorkload,
+		},
+		{
+			Name: "Get the deleted workload",
+			Command: *it.NewCommandLine(
+				"tanzu", "apps", "workload", "get", "test-create-git-annotations-workload", namespaceFlag),
+			ExpectedCommandLineError: true,
+			SkipClusterValidation:    true,
```
shashwathi, line 69: why is `SkipClusterValidation` set to true in this test? 
CRD and namespaces are present but workload is not.

testing/e2e/suite/comand_line_integration.go
```diff
@@ -0,0 +1,215 @@
...
+		}
+
+		if cl.CheckCommandOutput {
+			if diff := cmp.Diff(cl.ExpectedCommandLineOutput, cl.Command.GetOutput(), dfOpts...); diff != "" {
+				t.Errorf("%s\n(-expected, +actual)\n%s", cl.Name, diff)
+				t.FailNow()
+			}
+		}
+
+		if !cl.SkipClusterValidation {
```
shashwathi, line 147: `SkipClusterValidation` are there any cases we want to skip this for some reason? I thought  cluster validation would be part of setup for e2e tests like check CRDs exist and test ns exist

testing/e2e/suite/comand_line_integration.go
```diff
@@ -0,0 +1,215 @@
...
+					t.FailNow()
+				}
+
+			}
+		}
+		time.Sleep(time.Duration(cl.SleepMillisAfter) * time.Millisecond)
+	})
+}
+
+func preapreEnvironment(t *testing.T) {
```
shashwathi, line 187: typo `preapreEnvironment`

testing/e2e/test_cases/workload_create_test.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+func TestCreateFromGitWithAnnotations(t *testing.T) {
+	testSuite := it.CommandLineIntegrationTestSuite{
+		{
+			Name:         "Create workload with valid name from a git repo and annotations",
+			WorkloadName: "test-create-git-annotations-workload",
+			Command: func() it.CommandLine {
+				c := *it.NewCommandLine(
+					"tanzu", "apps", "workload", "create", "test-create-git-annotations-workload",
```
shashwathi, line 37: `tanzu apps` will default in all the commands so `NewCommandLine` can always prepend this as part of the function call. Less function args in that case.  If you end up doing this then rename func name to `NewTanzuAppsCommandLine`

testing/e2e/test_cases/workload_create_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/vmware-tanzu/apps-cli-plugin/testing/e2e/helpers"
+	it "github.com/vmware-tanzu/apps-cli-plugin/testing/e2e/suite"
+)
+
+func TestCreateFromGitWithAnnotations(t *testing.T) {
+	testSuite := it.CommandLineIntegrationTestSuite{
+		{
+			Name:         "Create workload with valid name from a git repo and annotations",
+			WorkloadName: "test-create-git-annotations-workload",
```
shashwathi, line 34: `WorkloadName` seems redundant. Better way to approach this would be to have ExpectedObject  and this object name could be used by cli client to fetch the actual object from API.

odinnordico: The expected object will have no name when validating non existing workload

testing/e2e/test_cases/workload_create_test.go
odinnordico: It was to avoid long files but yes it is easier to have it in the same one, the option to load from file remains but changed to create the wld inline

testing/e2e/test_cases/workload_create_test.go
odinnordico: added the new function `NewTanzuAppsCommandLine`

testing/e2e/suite/comand_line_integration.go
```diff
@@ -0,0 +1,215 @@
+//go:build integration
```
atmandhol, line 1: Type in filename. should be `command_line_integration.go` ?

odinnordico: correct, fixed

testing/e2e/suite/comand_line_integration.go
odinnordico: fixed

testing/e2e/suite/command_line_integration.go
```diff
@@ -0,0 +1,215 @@
...
+	Focus                     bool
+	WorkloadName              string
+	Command                   CommandLine
+	ExpectedCommandLineError  bool
+	ExpectedCommandLineOutput string
+	SkipClusterValidation     bool
+	ExpectedRemoteObject      client.Object
+	IsList                    bool
+	ExpectedRemoteList        client.ObjectList
+	SleepMillisBefore         int
```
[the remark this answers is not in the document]
odinnordico: some tasks may take some time to be processed by the cluster before assert, sleep time before can be removed

testing/e2e/suite/command_line_integration.go
```diff
@@ -0,0 +1,211 @@
...
+	time.Sleep(100 * time.Millisecond)
+	// create cluster supply chain
+	createClusterCmd := NewCommandLine("kubectl", "apply", "-f", "testdata/prereq/cluster-supply-chain.yaml")
+	err = createClusterCmd.Exec()
+	if err != nil {
+		t.Errorf("unexpected error: %v\n%s\n%s", err, createClusterCmd.GetOutput(), createClusterCmd.GetError())
+		cleanUp(t, true)
+		t.FailNow()
+	}
+	time.Sleep(100 * time.Millisecond)
```
shashwathi, line 200: Exec is a blocking call so remove sleep

testing/e2e/suite/command_line_integration.go
```diff
@@ -0,0 +1,211 @@
...
+
+func prepareEnvironment(t *testing.T) {
+	// create namespace
+	createNsCmd := NewCommandLine("kubectl", "create", "namespace", TestingNamespace)
+	err := createNsCmd.Exec()
+	if err != nil {
+		t.Errorf("unexpected error: %v\n%s\n%s", err, createNsCmd.GetOutput(), createNsCmd.GetError())
+		t.FailNow()
+	}
+	time.Sleep(100 * time.Millisecond)
```
shashwathi, line 191: Exec is a blocking call so remove sleep

shashwathi (approved):

PR #220, opened 2022-07-14
Allow users to pass CA self signed certs to push images to private registry

### LINKED ISSUE #116: Add flags to support --local-path workflow that allows user to override registry opts defaults
### Description of problem

Multiple instances were reported where a corporate firewall scrambles the SSL handshake between a dev's local machine and the external signed registry like azurecr.io. This results in apps plugin failing to push an image and an error message `certificate signed by unknown authority`. 

### Proposed solution

Currently `apps` plugin does not support flags to set the `CACertPaths`, `VerifyCerts` and `Insecure` [registry options](https://pkg.go.dev/github.com/k14s/imgpkg@v0.17.0/pkg/imgpkg/registry#Opts) through flags and VerifyCerts defaults to `true`. We should add flags to support boolean `VerifyCerts` and `Insecure` options, and default to secure options, but allow the user to override it using flags if they are behind firewall and run into this issue.

For Future state, we should provide a `config` option (once #108 is implemented) to let user change the defaults to insecure mode as they will always hit this firewall issue, but won't have to provide flags all the time.

CC @danfein @heyjcollins for design and product input regarding flag names.

Thanks to @cpage-pivotal for bringing this to our attention.

## Updates

We have decided to not implement support for overriding the following options
- `insecure` (will always default to false)
- `verifyCerts` (will always default to true)

Flags that will be added as part of this issue to support passing the following information
- `username`
- `password`
- `token`
- `caCertsPath`

Reference: https://pkg.go.dev/github.com/vmware-tanzu/carvel-imgpkg@v0.29.0/pkg/imgpkg/registry#Opts

warango4:
Pull request

Add new flags (`--registry-username`, `--registry-password`, `--registry-token`, `--registry-ca-cert`) to allow user pass self signed certs and push to a private registry. These flags come from `imgpkg`.

Fixes #116 

- Created binaries to test in Windows machine. Successfully pushed to a private Harbor registry.
- Used MacOS machine to push to same Harbor registry, worked too.
- Unit testing.

Signed-off-by: Wendy Arango warango@vmware.com


warango4:
image

warango4:
image

danfein (approved):

shashwathi (commented):
Registry options needs to be used in conjuction with local source image so can we add validation to ensure when any registry flags are passed `local-path` + `source-image` is provided as well.

docs/command-reference.md
```diff
@@ -1,6 +1,6 @@
 # Command reference
 
-- [Workload](command-reference/tanzu-apps-workload.md)
+- [Workload](command-reference/tanzu_apps_workload.md)
```
shashwathi, line 3: make a separate PR for doc name changes

warango4: I changed this here because I put some docs in the commands reference folder and when I ran `make docs`to add the new flags, all of them were erased. So I tried to correct at least these two here.

pkg/commands/workload.go
```diff
@@ -156,6 +161,15 @@ func (opts *WorkloadOptions) Validate(ctx context.Context) validation.FieldError
 		errs = errs.Also(validation.CompareQuantity(opts.LimitMemory, opts.RequestMemory, flags.RequestMemoryFlagName))
 	}
 
+	if opts.RegistryPassword != "" || opts.RegistryUsername != "" || opts.RegistryToken != "" {
```
shashwathi, line 164: why isnt this true for CA cert paths flag?

warango4: I missed it. Thanks for pointing it out.

shashwathi (approved):

rashedkvm (approved):
lgtm

PR #238, opened 2022-07-25
chore: Add color to healthy column

### LINKED ISSUE #227: Color code the health / status information in the supply chain section of workload get
### Description of problem

As the workload get output gets longer, the risk of folks overlooking important information increases. 
Adding emphasis / visual hierarchy via styling can help call attention to important/helpful bits of information.

### Proposed solution

<Please describe your proposed solution, preferably in the following style:>

```gherkin
Given I check the workload status via workload get
When something is stuck in the supply chain
Then the thing that is stuck should have extra attention called to it
```

### Example

We can use color to indicate good/bad states, making it easy to see at a glance when something is amiss.

for 
Ready column values    : true=green, false=red, unknown=grey
Healthy column values : true=green, false=red, unknown=grey

If ready is not always an accurate indicator of workload health, we could skip colorizing those values to prevent false alarms. 
ie, if something can be ready=true, but not up-and-running, the green would be misleading. Likewise, if it is possible for something to be not-ready in an expected state, the red might be a false alarm.

Ready column values    : true=white, false=white
Healthy column values : true=green, false=red

warango4:

Pull request

Add color to `HEALTHY` column in workload get output so it helps users to better visualize and distinguish the status of every step of their workload. 

Fixes #227 

Color code is
`true`=green/good
`false`=red/bad
`unknown`=cyan

### DIFF
--- a/pkg/cli-runtime/printer/format.go
+++ b/pkg/cli-runtime/printer/format.go
@@ -57,6 +57,17 @@ func ConditionStatus(cond *metav1.Condition) string {
 	}
 }
 
+func ColorConditionStatus(condStatus string) string {
+	switch condStatus {
+	case "True", "true":
+		return Ssuccessf(condStatus)
+	case "False", "false":
+		return Serrorf(condStatus)
+	default:
+		return Sinfof(condStatus)
+	}
+}
+
 func Labels(labelMap map[string]string) string {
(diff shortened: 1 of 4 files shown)


warango4:
image

warango4:
image

warango4:
image

shashwathi (approved):

PR #258, opened 2022-08-10
Added delivery on get output

### LINKED ISSUE #251: workload get includes deliverable information when available
### Description of problem

Workload get includes information about the supply chain process, but does not surface the delivery process information.

The supply chain information can help troubleshoot stuck deployments, and including the delivery information can be part of telling the user the story of the exiting journey thru the the machinery of TAP. 

### Proposed solution

```gherkin
Given I `get` a workload
When the information returns
Then I should see a table with the cluster delivery information
```

### Example

```
Supply chain  
   name:   source-to-url

   RESOURCE           READY   HEALTHY   TIME     OUTPUT  
   Source Provider    True    True      9m ago   ImageRepository/web3
   Image Builder      True    True      8m ago   Deliverable/web3
   Config Provider    True    True      7m ago   Image/web3
   App Config         True    True      6m ago   not found
   Config Writer      True    True      6m ago   not found
 
Delivery 
   name:   delivery-basic

   RESOURCE           READY   HEALTHY   TIME     OUTPUT
   delivery-source    True    True      9m ago   not found
   source-scanner     True    True      8m ago   not found
   source             True    True      7m ago   not found
   testing-pipeline   True    True      6m ago   not found

Messages
   No Messages Found

```

On an **iterate** cluster, the supply chain and delivery information should be available, and both tables can be shown.
On a **build** cluster, the supply chain info would be available, and the delivery table would be omitted
On a **run** cluster, the delivery information would be  available, and the supply table would be omitted

The **messages** section should have messages consolidated from both the workload and deliverable objects.

Combining supply and delivery steps in a single table. 
* saves a few lines in the workload-get output
* talking to @atmandhol it sounds like it would add complexity to the situations where you would only be able to see either supply or delivery stages, and when outputting yaml.

```
Deployment  
   Supply chain:   source-to-url
   Delivery chain: delivery-basic

   RESOURCE           READY   HEALTH   TIME      OUTPUT
   Source Provider    True    True     9m ago    <value>
   Image Builder      True    True     8m ago    <value>
   Config Provider    True    True     7m ago    <value>
   App Config         True    True     6m ago    <value>
   Config Writer      True    True     6m ago    <value>
   Delivery Source    True    True     6m ago    <value>
   Source Scanner     True    True     6m ago    <value>
   Source             True    True     6m ago    <value>
   Testing Pipeline   True    True     6m ago    <value>

Messages
   No Messages Found


```

odinnordico:
# Pull request

- Adds **_Delivery_** section to get output
  - Adds the Delivery Name
  - Adds the Deliverable resources with status information
-  Removes _last update_ and _ready_ from **_Supply Chain_** section
- Merge workload and delivery messages

Fixes #251 

Create different workloads in a TAP1.2 cluster and run _get_ command from apps plugin

image

image


pkg/commands/workload_get.go
```diff
@@ -160,22 +160,52 @@ func (opts *WorkloadGetOptions) Exec(ctx context.Context, c *cli.Config) error {
...
 
+	// Deliverable
+	c.Printf("\n")
+	c.Boldf("Delivery\n")
+	// Print workload deliverable resources
+	wldDeliverable := getWorkloadResourceByKind(workload, cartov1alpha1.WorkloadDeliverableResourceKind)
+	var deliverableStatusReadyCond *metav1.Condition
+	notFoundMsg := printer.AddPaddingStart("Delivery resources not found.\n")
+	deliverable := &cartov1alpha1.Deliverable{}
+	if wldDeliverable == nil || wldDeliverable.StampedRef == nil {
```
shashwathi, line 178: `wldDeliverable.StampedRef` is checked twice. Once in `getWorkloadResourceByKind` func and here as well.

pkg/apis/cartographer/v1alpha1/deliverable.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=deliverables,categories=all,shortName=dlv
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Source",type="string",JSONPath=`.spec.source['git.url','image']`
+// +kubebuilder:printcolumn:name="Delivery",type="string",JSONPath=".status.deliveryRef.name"
+// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=`.status.conditions[?(@.type=='Ready')].status`
```
shashwathi, line 26: you could remove the kubebuilder tags and the description of fields from the file. This is a copy and not the source of CRDs location so you dont need make it descriptive

pkg/commands/workload_get.go
```diff
@@ -160,22 +160,52 @@ func (opts *WorkloadGetOptions) Exec(ctx context.Context, c *cli.Config) error {
...
+		c.Infof(notFoundMsg)
+	} else {
+		if err := c.Get(ctx, client.ObjectKey{Namespace: wldDeliverable.StampedRef.Namespace, Name: wldDeliverable.StampedRef.Name}, deliverable); err != nil {
```
shashwathi, line 181: if Get fails with an error, `deliverable` is nil and since code path does not return here and in the section `deliverable.Status` is accessed so this could end up with nil pointer exception.

Return the error here or check the object not nil before accessing it

pkg/apis/cartographer/v1alpha1/deliverable.go
odinnordico: Aren't those required to generate deep copy?

pkg/commands/workload_get.go
shashwathi, line 178: swap if and else condition
- focus the main logic which is usual case and then the exception

shashwathi (commented):
looks good

- there is lot of repeated code for creating dies obj for deliverable conditions and other parts of  the Deliverable object. Revisit that logic and see if you can create func in Die package or declare those variable outside of test to be reused.

pkg/apis/cartographer/v1alpha1/deliverable.go
shashwathi: CLi uses controller gen to generate fakes. I see that other CRDs have kube builder tags as well. I will create a separate story to clean it up.

pkg/commands/workload_get.go
odinnordico: Removed from `if`

pkg/commands/workload_get.go
odinnordico: Null object checked

pkg/commands/workload_get.go
odinnordico: Done

shashwathi (approved):

PR #261, opened 2022-08-11
feat: removed deliverable of supplychain table

### LINKED ISSUE #257: Remove "deliverable" resource from the supply chain table displayed in `workload get` output
### Description of problem

The deliverable resource that's included in the "Supply Chain" resource table displayed in the output from `workload get` doesn't provide much value to users and should be removed.

1. the HealthRule for the deliverable is _always_ True (won't show False even if it's failed) - so not much actionable value can be had from it's inclusion as-is
   - in addition, the meaning/purpose of the deliverable step is not intuitively obvious to users and can cause confusion
1. as part of (#251), we will be adding the deliverable info in workload get as part of the (new/pending) `clusterDelivery` section, and messages from the deliverable will be displayed in the  `Messages` section and this additional content should account from the removal of deliverable from the supply chain table.
1.   The deliverable resource that's displayued in the "Supply Chain" section of `workload get` for a workload which was selected by part of the OOTB `source-to-url` supply chain 

### Proposed solution

remove the deliverable row from the supply chain table

```gherkin
Given my workload X has been deployed successfully to my target cluster via the `source-to-url` supply chain
When I run "tanzu apps workload get X"
Then I DO NOT see a `deliverable` resource row displayed in the Supply Chain resources table in the command output
```

### Example

```
ta get X
---
# X: Ready
---
...

Supply Chain
   name:          source-to-url
   last update:   42h
   ready:         True

   RESOURCE          READY   HEALTHY   TIME   OUTPUT
   source-provider   True    True      42h    GitRepository/app4
   image-builder     True    True      42h    Image/app4
   config-provider   True    True      42h    PodIntent/app4
   app-config        True    True      42h    ConfigMap/app4
   config-writer     True    True      42h    Runnable/app4-config-writer

   Messages
   No messages found.

...

To see logs: "tanzu apps workload tail X"
```

Leaving the deliverable row as-is

### Additional context
#251 should be implemented prior to this issue.

odinnordico:
Pull request

Removes `Deliverable` row from supply chain resources in `get` output

Fixes #257 

Local install of apps plugin against a Tap1.2 cluster getting the same `Workload`

image

### DIFF
--- a/pkg/printer/helpers.go
+++ b/pkg/printer/helpers.go
@@ -16,10 +16,20 @@ limitations under the License.
 
 package printer
 
-import "strings"
+import (
+	"strings"
+
+	cartov1alpha1 "github.com/vmware-tanzu/apps-cli-plugin/pkg/apis/cartographer/v1alpha1"
+)
 
 const (
 	paddingStart = 3
+)
+
+var (
+	supplyChainResourcesKindExcludeList = map[string]bool{
+		cartov1alpha1.DeliverableKind: true,
+	}
 )
 
 func AddPaddingStart(text string) string {
--- a/docs/commands-details/workload_get.md
+++ b/docs/commands-details/workload_get.md
@@ -35,7 +35,6 @@ Supply Chain
 
    RESOURCE          READY   HEALTHY   TIME    OUTPUT
    source-provider   True    True      3m51s   ImageRepository/rmq-sample-app
-   deliverable       True    True      3m55s   Deliverable/rmq-sample-app
    image-builder     True    True      101s    Image/rmq-sample-app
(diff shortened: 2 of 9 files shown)


pkg/printer/helpers.go
```diff
@@ -22,6 +22,12 @@ const (
 	paddingStart = 3
 )
 
+var (
+	supplyChainResourcesKindExcludeList = map[string]bool{
+		"Deliverable": true,
```
shashwathi, line 27: use const for `Deliverable` kind

odinnordico: `DeliverableKind` constant used

shashwathi (approved):

PR #282, opened 2022-08-23
fix: remove TANZU_APPS_SOURCE_IMAGE from allowed

odinnordico:
# Pull request

Removes TANZU_APPS_SOURCE_IMAGE environment variable from allowed to use as default

Fixes #

### DIFF
--- a/pkg/flags/env_var.go
+++ b/pkg/flags/env_var.go
@@ -25,7 +25,6 @@ var (
 		FlagToEnvVar(RegistryPasswordFlagName): true,
 		FlagToEnvVar(RegistryTokenFlagName):    true,
 		FlagToEnvVar(RegistryUsernameFlagName): true,
-		FlagToEnvVar(SourceImageFlagName):      true,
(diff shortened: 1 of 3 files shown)


pkg/flags/env_var.go
```diff
@@ -25,7 +25,7 @@ var (
 		FlagToEnvVar(RegistryPasswordFlagName): true,
 		FlagToEnvVar(RegistryTokenFlagName):    true,
 		FlagToEnvVar(RegistryUsernameFlagName): true,
-		FlagToEnvVar(SourceImageFlagName):      true,
+		FlagToEnvVar(SourceImageFlagName):      false,
```
shashwathi, line 28: can you remove this line

shashwathi (changes requested):

pkg/flags/env_var.go
odinnordico: Removed

shashwathi (approved):

PR #294, opened 2022-08-25
fix: validate namespace when validating workload

### LINKED ISSUE #271: Source image is uploaded when namespace is invalid for workload creation/apply

### Describe the bug
When user creates/applies workload from local-path and source-image flag in some non existent namesspace, source code/jar is uploaded to registry and then workload creation fails because namespace does not exist.

User has to setup namespace and then re-run the command but source image upload has already been done in the previous step.

### Expected behavior
Fail before source image upload because namespace does not exist. Preliminary name and namespace check should be done before triggering source image upload.

### Steps to Reproduce

```
tanzu apps workload create pet-img-5 --source-image gcr.io/something/helloworldtestjar:source --local-path pkg/source/testdata/helloworld.jar -n notexist

? Publish source in "pkg/source/testdata/helloworld.jar" to "gcr.io/something/helloworldtestjar:source"? It may be visible to others who can pull images from that repository Yes
Publishing source in "pkg/source/testdata/helloworld.jar" to "gcr.io/something/helloworldtestjar:source"...
Published source
Create workload:
      1 + |---
      3 + |kind: Workload
      4 + |metadata:
      5 + |  name: pet-img-5
      6 + |  namespace: notexist
      7 + |spec:
      8 + |  source:
      9 + |    image: gcr.io/something/helloworldtestjar:source@sha256:980ab26419c9139b973daf1f2adc1ebacf91fcfd416769d03dd09768c12a1f1b

? Do you want to create this workload? Yes
Error: namespaces "notexist" not found
Error: exit status 1

✖  exit status 1 
```

```sh
v0.8.0
```

odinnordico:
# Pull request

Validates if the provided namespace can be accessed by the user when validating if workload exists.

Fixes #271 

Loca installation of plugin against TAP1.2 cluster in GKE

image

Try to create a workload from source code in the default namespace and a non-existing namespace.

This validation applies to all the `apply`/`create` subcommands, if the namespace does not exist or is not accessible for the user it will throw an error

### DIFF
--- a/pkg/commands/workload_create_test.go
+++ b/pkg/commands/workload_create_test.go
@@ -23,6 +23,7 @@ import (
 	"testing"
 	"time"
 
+	diecorev1 "dies.dev/apis/core/v1"
 	diemetav1 "dies.dev/apis/meta/v1"
 	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/mock"
@@ -98,8 +99,16 @@ func TestWorkloadCreateCommand(t *testing.T) {
 
 	scheme := runtime.NewScheme()
 	_ = cartov1alpha1.AddToScheme(scheme)
+	_ = corev1.AddToScheme(scheme)
 
 	var cmd *cobra.Command
+
+	givenNamespaceDefault := []client.Object{
+		diecorev1.NamespaceBlank.
+			MetadataDie(func(d *diemetav1.ObjectMetaDie) {
+				d.Name(defaultNamespace)
+			}),
+	}
 
 	table := clitesting.CommandTestSuite{
 		{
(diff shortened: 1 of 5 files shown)


pkg/commands/workload_create_test.go
```diff
@@ -110,10 +112,22 @@ func TestWorkloadCreateCommand(t *testing.T) {
 		{
 			Name: "no source",
 			Args: []string{workloadName},
+			GivenObjects: []client.Object{
+				diecorev1.NamespaceBlank.
```
shashwathi, line 116: you can define it as var outside the test table and use that in test. 
```
defaultNamepace := diecorev1.NamespaceBlank.
					MetadataDie(func(d *diemetav1.ObjectMetaDie) {
						d.Name(defaultNamespace)
					})
```

shashwathi (changes requested):

shashwathi (approved):

PR #333, opened 2022-09-20
fixes test cases for wait file

### LINKED ISSUE #332: Unit test failing in Windows OS for cli runtime wait.go

### Describe the bug
Unit test failing in Windows OS for wait.go file test cases
OS: Windows
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
    	panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x112c7a2]

### Expected behavior
All unit test cases are successful 

### Steps to Reproduce
clone apps-cli-plugin code git clone https://github.com/vmware-tanzu/apps-cli-plugin in windows and run go test ./....
```sh
```

shaheerkootteeri:
Pull request

fixes test cases for wait file which was erroring when testing in windows OS

Fixes #332 

Validated unit test cases in mac as well as windows.

### DIFF
--- a/pkg/cli-runtime/wait/wait_test.go
+++ b/pkg/cli-runtime/wait/wait_test.go
@@ -158,7 +158,7 @@ func TestUntilDelete(t *testing.T) {
 	defer func() {
 		BackOffTime = previousBackOffTime
 	}()
-	BackOffTime = 10 * time.Millisecond
+	BackOffTime = 30 * time.Millisecond
 	workload := &cartov1alpha1.Workload{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: defaultNamespace,
@@ -189,7 +189,7 @@ func TestUntilDelete(t *testing.T) {
 		reactorFunc clitesting.ReactionFunc
 	}{{
 		name:    "smaller timeout",
-		timeout: time.Millisecond,
+		timeout: time.Nanosecond,
 		err:     context.DeadlineExceeded,
 	}, {
 		name:    "API not found error",


pkg/cli-runtime/wait/wait_test.go
```diff
@@ -191,6 +191,13 @@ func TestUntilDelete(t *testing.T) {
 		name:    "smaller timeout",
 		timeout: time.Millisecond,
 		err:     context.DeadlineExceeded,
+		reactorFunc: func(action clitesting.Action) (bool, runtime.Object, error) {
```
shashwathi, line 194: this test case should error out when context being done and not invoke get call. This change is altering how `context.DeadlineExceeded` error is generated.
If this test fails in windows, i would reduce the timeout nanosecond and see if that helps so it doesnt trigger the 2nd condition of the ticker

shaheerkootteeri: changing that to Nano seconds are not working as it is expecting to have the reactorFunction.

pkg/cli-runtime/wait/wait_test.go
```diff
@@ -158,7 +158,7 @@ func TestUntilDelete(t *testing.T) {
 	defer func() {
 		BackOffTime = previousBackOffTime
 	}()
-	BackOffTime = 10 * time.Millisecond
+	BackOffTime = 30 * time.Millisecond
```
shashwathi, line 161: why not pass `BackOffTime` as func function parameter?

shashwathi (changes requested):

pkg/cli-runtime/wait/wait_test.go
shaheerkootteeri: Do you mean adding Backofftime as a function parameter to UntilDelete ? 
Having the BackOfftime public enables calling function to reset the value from default.

odinnordico (approved):

warango4 (approved):

pkg/cli-runtime/wait/wait_test.go
shashwathi: overriding public variable is not great way to set a variable. In future to add tests can you see how this will become a problem?

shaheerkootteeri: workload delete uses the untilDelete function in wait file , and if we remove this public variable and have it accept as a function parameter, we will need to have prepare for each test case that store the backOfftime in context for each of test case and in workload delete we pull the value back from context or default to 5 second.

shashwathi (approved):

PR #358, opened 2022-09-30
fixes paths in workload test file to use filepath library

shaheerkootteeri:
Pull request

PR fixes the paths in the workload test file to use filepath library to make OS agnostic 

Fixes #

Tested unit test cases in both Mac as well as windows machine

### DIFF
--- a/pkg/commands/workload_test.go
+++ b/pkg/commands/workload_test.go
@@ -55,6 +55,8 @@ import (
 	"github.com/vmware-tanzu/apps-cli-plugin/pkg/source"
 )
 
+var localSource = filepath.Join("testdata", "local-source")
+
 func TestWorkloadCommand(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/commands/workload_test.go
```diff
@@ -73,6 +76,10 @@ func TestWorkloadCommand(t *testing.T) {
 }
 
 func TestWorkloadOptionsValidate(t *testing.T) {
+	localRepo := filepath.Join(string(filepath.Separator), "path", "to", "local", "repo")
```
odinnordico, line 79: can `string(filepath.Separator)` be a global const/var to not cal `string(...)` each time?

odinnordico (changes requested):

shashwathi (approved):

odinnordico (dismissed):

shashwathi (approved):

PR #385, opened 2022-10-18
Update tail help message

### LINKED ISSUE #369: add `--since 1h --timestamp` to the command tip displayed in `workload create/apply/update/get`
### Description of problem
In the output from `workload create/apply/get`, we recommend users run `workload tail` to view logs for their workload via a tip at the very end of the command output.

In a troubleshooting scenario, a problem has already occurred, so running the recommended `tail` command isn't likely to return log messages related to that problem.

We've received feedback from users asking how to view logs that were emitted prior to the moment the `workload tail` command is invoked. This indicates users aren't readily discovering that this feature is already supported.
And the fact that we're recommending a `tail` incantation which doesn't improve a user's ability to discover the feature while also not being likely to provide value as-is, is a UX miss.

We've also observed many users tail workload logs and include `--timestamp`.
Additionally, log message timestamps are considered valuable and we should include the flag in our recco.

### Proposed solution

Include `--timestamp --since 1h` in the example command being recommended. 

### Example

Update the command tip from: 
```To see logs:   "tanzu apps workload tail foo"```
To:
```To see logs:   "tanzu apps workload tail foo --timestamp --since 1h"```

recommend use of `--since` outside of the recommended command: 
```To see logs:   "tanzu apps workload tail foo" (include "--since" to view previous logs)```

warango4:
Pull request

Update `tail` help message in workload create/apply/get so it will tell users to use `timestamp` and `since` flags 

Fixes #369 

- Updated unit testing

### DIFF
--- a/docs/commands-details/workload_create_update_apply.md
+++ b/docs/commands-details/workload_create_update_apply.md
@@ -30,7 +30,7 @@ Create workload:
 ? Do you want to create this workload? Yes
 Created workload "pet-clinic"
 
-To see logs:   "tanzu apps workload tail pet-clinic"
+To see logs:   "tanzu apps workload tail pet-clinic --timestamp --since 1h"
 To get status: "tanzu apps workload get pet-clinic"
 
 ```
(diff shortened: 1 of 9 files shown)


odinnordico (approved):
LGTM

shaheerkootteeri (approved):

PR #395, opened 2022-10-26
feat: Allow users to pass url for --file flag

### LINKED ISSUE #372: Support remote url resources with -f flag
### Description of problem

`kubectl` users often point to resources on the internet using their url while using the `-f` flag. 

e.g. 
```
kubectl apply -f https://raw.githubusercontent.com/.../.../workload.yaml
```

Currently, tanzu apps CLI only supports pointing to files in the user's local machine using the `-f` flag. 

### Proposed solution

We should include the capability to allow users to point at remote urls.

```gherkin
Given I have a file that is accessible using a URL
When I run tanzu apps workload create/apply using -f pointing to that URL
Then CLI should be able to read that file and create a workload
```

warango4:
Pull request

Add the capability to use remote file (URL) to create/apply a workload

Fixes #372 

- Updated unit test
- Updated e2e test

### DIFF
--- a/pkg/commands/workload.go
+++ b/pkg/commands/workload.go
@@ -23,6 +23,8 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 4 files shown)


pkg/commands/workload.go
```diff
@@ -616,21 +618,56 @@ func (opts *WorkloadOptions) Create(ctx context.Context, c *cli.Config, workload
...
-	f, err := os.Open(opts.FilePath)
-	in = f
-	if f == nil && opts.FilePath == "-" {
+	isURL, err := isUrl(opts.FilePath)
+	if err != nil {
+		return fmt.Errorf("unable to check if filepath %q is a valid url: %w", opts.FilePath, err)
+	}
+	switch {
+	case isURL:
+		in, err = opts.getUrlFileContent()
```
odinnordico, line 627: **Question:** `isUrl(string)` returns `true` when the url points to a resource from a`ftp`, `smb`, etc (i.e. `ftp://myserver/my-sub/my-workload.yaml`) server, does `WorkloadOptions.getUrlFileContent()` suports retrieving content usingthosep protocols?

pkg/commands/workload.go
```diff
@@ -616,21 +618,56 @@ func (opts *WorkloadOptions) Create(ctx context.Context, c *cli.Config, workload
...
+		if err != nil {
+			return fmt.Errorf("unable to read file from url %q: %w", opts.FilePath, err)
```
odinnordico, line 629: please @danfein check this message, i think using _resource_ instead of _file_ is better suited

pkg/commands/workload.go
```diff
@@ -616,21 +618,56 @@ func (opts *WorkloadOptions) Create(ctx context.Context, c *cli.Config, workload
...
 	return nil
 }
 
+func (opts *WorkloadOptions) getUrlFileContent() (io.Reader, error) {
+	resp, err := http.Get(opts.FilePath)
+	if err != nil {
+		return nil, err
+	}
+
+	body, err := io.ReadAll(resp.Body)
```
odinnordico, line 654: **Question:** Why err is not validated here and returned if not `nil`?

pkg/commands/workload.go
atmandhol: Let's plan to support just http/https for the scope of this GH issue. We can add support for other protocols in the future if there is an ask for it.

pkg/commands/workload.go
danfein: Talked with @warango4  - suggested simplifying down to `unable to read from url <value>` 

Shortens the message, and sidesteps the question of what to call the thing we are reading. :)

odinnordico (approved):
LGTM

shaheerkootteeri (dismissed):
LGTM

anibmurthy (dismissed):
lgtm

pkg/commands/workload.go
```diff
@@ -616,21 +618,61 @@ func (opts *WorkloadOptions) Create(ctx context.Context, c *cli.Config, workload
...
+	if err != nil {
+		return fmt.Errorf("unable to check if filepath %q is a valid url: %w", opts.FilePath, err)
+	}
+	switch {
+	case isURL:
+		in, err = opts.getUrlFileContent()
+		if err != nil {
+			return fmt.Errorf("unable to read from url %q: %w", opts.FilePath, err)
+		}
+	case opts.FilePath == "-":
```
anibmurthy, line 631: **Suggestion**: It is better to modify this to an `if-else` chain:

- as there is an immediate else-if continuation. And
- Case matching is not on the same variable

odinnordico (approved):

PR #402, opened 2022-11-03
feat: Add --update-strategy file to change apply behavior

### LINKED ISSUE #336: As an app dev I can pass a flag/value on `workload apply` to overwrite existing workload configuration on the target cluster with my local workload.yaml configuration
### Description of problem

We've made it easy to add or update fields/properties in the workload.yaml via `tanzu apps workload apply -f workload.yaml` but it's impossible to delete existing fields/properties (when removed from workload.yaml) via the same command.
Field deletion is only supported via flags currently (e.g. user who needs to remove envvar `foo: bar` must invoke `tanzu apps workload apply MY-APP --env foo-` 

Having to delete workload fields via direct invocation of Tanzu CLI apps plugin isn't intuitive for someone living/breathing in the IDE where they're used to invoking updates directly through the IDE interface itself where the workload.yaml is readily available to edit.

Today the apps plugin mutates/merges the workload object on the cluster when a dev `tanzu apps workload apply`s the workload.yaml on their local filesystem via `--file workload.yaml`. 

Originally, this was taken because we didn't want to assume the person invoking the change knew they had included all the fields required.
This may be helpful at times, but there are use cases where it adds significant friction (e.g. when using the IDE, the user has to hop out of their regular IDE-based flow, and invoke the apps CLI command directly). 

We need a way for the user to assert their workload definition as definitive and completely replace the version on-cluster.

### Proposed solution
Changing the behavior of the command introduces a breaking change so we must introduce a backward/forward-compatible means to replace workload configuration.
Introducing a flag/value enables end users to triggers the desired behavior without disrupting existing scripts/user-flows.

@doddatpivotal  suggested using  a `--force` flag which, when included, would tell the apps plugin to completely overwrite the existing workload object on the cluster with the workload.yaml on the developer's local filesystem (`tanzu apps workload apply -f workload.yaml --force`) - thank you for the suggestion Dodd :) 

Let's go with the spirit of Dodd's suggestion while optimizing for the ultimate goal of making "replace" the default behavior in the future.

Some ideas:
* ~~`--force`~~ (bool)
* ~~`--replace`~~ (bool)
* ~~`--always-overwrite`~~ (bool)
* ~~`--overwrite`~~ (bool)
* ~~`--file-merge`~~ (bool)
* `--update-strategy` (merge/replace)

**`--update-strategy`** is flag we've decided to go with.

#### Requirements:
1. [x] New flag `--update-strategy` added for sub-command `apply` (values: `merge`(default)/`replace`)
1. [x] Autocompletion for `--update-strategy [TAB]` >> merge, replace
1. [x] Deprecation warning displayed in `apply` output when `--file ...` flag is included:
  * `❗ WARNING: Configuration file update strategy is changing. By default, provided configuration files will replace rather than merge existing configuration. The change will take place in the January 2024 TAP release (use "--update-strategy" to control strategy explicitly).` 
1. [x] must be inclusive of `--file workload.yaml` (should throw an error if `--file workload.yaml` isn't included)
1. [ ] **Docs** Release Note > Deprecation notice regarding default behavior change from `merge` to `replace` in January 2024 TAP release
1. [ ] **Docs update** apply command reference - include complete explanation of the new flag, include examples for `--update-strategy merge` and `--update-strategy replace` 
1. [x] help for `--update-strategy` included in `tanzu apps apply -h` output
  * `--update-strategy          specify configuration file update strategy (supported strategies: merge[default], replace )`
 1. [x] `--update-strategy` doesn't make sense for `tanzu apps workload create` so if it's possible to not include it in `create -h` that is strongly desired (although passing `--update-strategy` would be a no-op either way so if it's technically impossible or exceptionally difficult to exclude it from `workload create` it would be sub-optimal, but not a deal-breaker.
1. [x] TO CONSIDER - read value from ENVVAR, when exists (name of envvar: `TANZU_APPS_UPDATE_STRATEGY`)

### Acceptance Criteria
```gherkin
Given I've created a workload that includes an environment variable "foo=bar"
When I remove
"
env:
  -name: foo
   value: bar
" 
from the workload.yaml on my filesystem and invoke "tanzu apps workload apply -f workload.yaml --update-strategy replace"
Then the workload object on the cluster is updated so it matches my local workload.yaml exactly (envvar foo=bar has been removed)
```

And make sure there's no regression WRT uploading updated source code and also passing in the workload.yaml
```gherkin
Given I've created a workload from local source that includes an environment variable "foo=bar"
e.g. "tanzu apps workload apply my-app --local-path . --source-image gcr.io/pivotal-knative/jc-apps --type web --env foo=bar --yes"
When I run "tanzu apps workload get my-app --export > workload.yaml" to save the current yaml to my local filesystem,
AND edit the workload.yaml to remove
"
env:
  -name: foo
   value: bar
"
AND make and save a change to the source code for my-app on my local filesystem
AND invoke "tanzu apps workload apply -f workload.yaml --update-strategy replace"
Then I can see in the diff output that the apps plugin uploaded the new code and replaced the current "spec.source.image" value with the new digest for the uploaded code
AND that the "foo=bar" env var has also been removed 
```

And current merge behavior hasn't regressed
```gherkin
Given I've created a workload named "bar" from local source
When download the current workload definition from the cluster via "tanzu apps workload get bar --export > bar.yaml"
and I remove
"
  name: bar
  namespace: default
  ...
spec:
  source:
    image: gcr.io/pivotal-knative/jc-apps:latest@sha256:2b0bb7ae0d403a98fac94d58ecdebf3272c336c9b7ba02a2f9cbb84604f2b833
" 
and save that as my local workload.yaml on my filesystem and invoke 
"tanzu apps workload apply -f bar.yaml --update-strategy merge"
OR
"tanzu apps workload apply -f bar.yaml"
Then the apps plugin will not make an update because there are no changes detected 
```

And file validation hasn't regressed when `replace` strategy is invoked
```gherkin
Given I've created a workload "bar" from local source
When download the current workload definition from the cluster 
and I remove required "name" field 
"
  name: bar
  ...
from the workload.yaml on my filesystem and invoke "tanzu apps workload apply -f workload.yaml --update-strategy replace"
Then the apps plugin will return an error because required field is missing from the workload definition("name" ).
```

1. requiring all removals to be executed via flags (as exists today)
1. create a new sub-command `tanzu apps workload replace`
1. provide a mechanism to persist a configured default

warango4:
Pull request

This is an initial commit, the first part that is creating the flag, allow autocompletion and adding the warning message for this new breaking change. Following commits will introduce the behavior the program should have in case of using this flag (checks, errors)

Fixes #336 


warango4:
image

warango4:
Without setting `--update-strategy` (uses default as `merge`)

image

warango4:
Setting `--update-strategy` to `replace`

image

pkg/commands/workload_apply.go
```diff
@@ -112,9 +129,23 @@ func (opts *WorkloadApplyOptions) Exec(ctx context.Context, c *cli.Config) error
 
 		workload.Spec.MergeServiceAccountName(serviceAccountCopy)
 	}
+	if opts.UpdateStrategy == MergeUpdateStrategy {
+		workload.Merge(fileWorkload)
+	}
 
-	workload.Merge(fileWorkload)
+	if opts.UpdateStrategy == ReplaceUpdateStrategy {
```
warango4, line 136: When trying to assign 
`workload = fileWorkload`

all the object metadata got lost so it threw an error for some fields that the system automatically populates. So this approach tries to first assign the metadata that comes from the workload that's currently in the cluster and then replace `labels` and `annotations` according to the user changes in the file.

warango4: Some other way that I tried and works as well is trying to just add the system populated fields to the workload metadata after it has been replaced with the workload from file

```
        workload = fileWorkload
        if currentWorkload != nil {
            if !reflect.DeepEqual(currentWorkload.ObjectMeta, (metav1.ObjectMeta{})) {
                wldMeta := workload.GetObjectMeta()
                currentWldMeta := currentWorkload.GetObjectMeta()

                wldMeta.SetResourceVersion(currentWldMeta.GetResourceVersion())
                wldMeta.SetUID(currentWldMeta.GetUID())
                wldMeta.SetGeneration(currentWldMeta.GetGeneration())
            }
        }
```

I didn't use this approach because, if eventually, there are other `n` fields like this, then we'll have to manually add them and it will become untenable in time

warango4: On the other hand, if user tries to update fileds like `OwnerReferences`, `Finalizers`, `ManageFields`, then these won't be updated, 'cause some logic to manage these is missing.

pkg/apis/cartographer/v1alpha1/workload_helpers.go
```diff
@@ -97,6 +97,33 @@ func (w *WorkloadSpec) MergeServiceAccountName(serviceAccountName string) {
 	w.ServiceAccountName = serviceAccountNamePtr
 }
 
+func (w *Workload) ReplaceMetadata(updates *Workload) {
+	w.ReplaceAnnotations(updates)
```
warango4, line 101: Should something like the following be added? 

```
	w.GenerateName = updates.GenerateName
	w.DeletionGracePeriodSeconds = updates.DeletionGracePeriodSeconds
	w.OwnerReferences = updates.OwnerReferences
	w.Finalizers = updates.Finalizers
	w.ManagedFields = updates.ManagedFields
```

If so, we'd be facing the same problem: if fields that user can manage start to increase, this is going to start very difficult to manage.

docs/command-reference/tanzu_apps_workload_apply.md
```diff
@@ -63,6 +63,7 @@ tanzu apps workload apply --file workload.yaml
       --tail                           show logs while waiting for workload to become ready
       --tail-timestamp                 show logs and add timestamp to each log line while waiting for workload to become ready
   -t, --type type                      distinguish workload type
+      --update-strategy string         specify configuration file update strategy (supported strategies: merge, replace) (default "merge")
```
heyjcollins, line 66: I'd recommend a slightly shorter version (adjust spacing as needed - I'm pretty sure I got it wrong):

--update-strategy   string          specify configuration file update strategy (supported strategies: merge[default], replace )

warango4: This is the way cobra is adding the help text

pkg/commands/workload_apply.go
warango4: As per discussion with @shashwathi, the decided approach is to copy all of the metadata as it comes and then set the system populated fields, since metadata is not likely to have big/non-manageable changes.

pkg/commands/testdata/replace-update-strategy/replace-annotations.yaml
```diff
@@ -0,0 +1,28 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: carto.run/v1alpha1
+kind: Workload
+metadata:
+  name: spring-petclinic
+  annotations:
```
heyjcollins, line 19: ## I'm not sure this is correct.
The workload object passes all annotation references to the supply chain via a param named "annotations"
And the supply chain converts the annotation param values to actual kubernetes annotations on the backend.

### Here's an example of a workload.yaml with some annotations that were added via the apps CLI plugin:

```
ta apply tanzu-java-web-app  --annotation autoscaling.knative.dev/min-scale=2 --annotation autoscaling.knative.dev/max-scale=5
🔎 Update workload:
...
  6,  6   |    apps.tanzu.vmware.com/workload-type: web
  7,  7   |  name: tanzu-java-web-app
  8,  8   |  namespace: ns1
  9,  9   |spec:
     10 + |  params:
     11 + |  - name: annotations
     12 + |    value:
     13 + |      autoscaling.knative.dev/max-scale: "5"
     14 + |      autoscaling.knative.dev/min-scale: "2"
...

ta get tanzu-java-web-app --export
---
apiVersion: carto.run/v1alpha1
kind: Workload
metadata:
  labels:
    apps.tanzu.vmware.com/workload-type: web
  name: tanzu-java-web-app
  namespace: ns1
spec:
  params:
  - name: annotations
    value:
      autoscaling.knative.dev/max-scale: "5"
      autoscaling.knative.dev/min-scale: "2"
  source:
    image: gcr.io/pivotal-knative/jc-apps:latest@sha256:26d0e12a3e6e1a1d74bf18d52f3d39aa5e9555a7b69b419e4684bd36b77aa749
```

### And if I inspect the running pod, you can see those have been transformed into true annotations (here's a snippet of the output):
```
k get pod tanzu-java-web-app-00002-deployment-7fdf4b8fbb-rhrfj
NAME                                                   READY   STATUS    RESTARTS   AGE
tanzu-java-web-app-00002-deployment-7fdf4b8fbb-rhrfj   2/2     Running   0          3m37s
11/16/22-16:22:43 ~ %k describe pod tanzu-java-web-app-00002-deployment-7fdf4b8fbb-rhrfj
Name:             tanzu-java-web-app-00002-deployment-7fdf4b8fbb-rhrfj
Namespace:        ns1
Priority:         0
Service Account:  default
Node:             gke-iterate-1-4-x-default-pool-61cc6a50-1cks/10.150.0.51
Start Time:       Wed, 16 Nov 2022 16:19:06 -0800
Labels:           app=tanzu-java-web-app-00002

                  ...

Annotations:      autoscaling.knative.dev/max-scale: 5
                  autoscaling.knative.dev/min-scale: 2
                 
                  ...
         
Events:
  Type     Reason     Age                    From               Message
  ----     ------     ----                   ----               -------
  Normal   Scheduled  3m48s                  default-scheduler  Successfully assigned ns1/tanzu-java-web-app-00002-deployment-7fdf4b8fbb-rhrfj to gke-iterate-1-4-x-default-pool-61cc6a50-1cks
  
  ...

  Warning  Unhealthy  3m41s (x3 over 3m43s)  kubelet            Readiness probe failed: Get "http://10.4.3.32:8012/": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

warango4: so what should happen if the user does something like this? Directly add an annotation into the workload metadata. Should it fail? 

Because in another test (refer to the integration test), these annotations are being passed through flags, and in the workload yaml they're added as params. Both ways are currently working.

testing/e2e/workload_test.go
```diff
@@ -516,6 +517,94 @@ func TestCreateFromGitWithAnnotations(t *testing.T) {
...
+							},
+						},
+					},
+					Params: []cartov1alpha1.Param{
+						{
+							Name:  "services",
+							Value: v1.JSON{Raw: []byte(`[{"image":"mysql:5.7","name":"mysql"},{"image":"postgres:9.6","name":"postgres"}]`)},
+						},
+						{
+							Name: "annotations",
```
warango4, line 599: @heyjcollins

docs/command-reference/tanzu_apps_workload_apply.md
heyjcollins: Assuming that means we can't do what I suggested so I'll resolve this conversation.

pkg/commands/testdata/replace-update-strategy/replace-annotations.yaml
warango4: @shashwathi any thoughts?

testing/e2e/workload_test.go
```diff
@@ -516,6 +517,94 @@ func TestCreateFromGitWithAnnotations(t *testing.T) {
 				return nil
 			},
 		},
+		{
+			Name:         "Update the created workload with `replace` update strategy",
+			WorkloadName: "spring-petclinic",
+			Command: func() it.CommandLine {
+				c := *it.NewTanzuAppsCommandLine(
+					"workload", "apply", "spring-petclinic",
```
shaheerkootteeri, line 525: Do we need a test case with no update strategy specified so that it picks up the default and do nothing or show no change in workload?

shaheerkootteeri:
Also is it expected to display the warning with the other errors as below 
image

shaheerkootteeri:
Do we need any Emojis for the message "workload is unchanged ?" is that message more occurring than before after the introduction of this update-strategy 
image

pkg/cli-runtime/testing/validatable_table.go
```diff
@@ -49,9 +49,21 @@ type ValidatableTestCase struct {
 
 	// ShouldValidate is true if the validatable object is valid
 	ShouldValidate bool
+
+	// lifecycle
+
+	// Prepare is called before the command is executed. It is intended to prepare that broader
+	// environment before the specific table record is executed. For example, changing the working
+	// directory or setting mock expectations.
+	Prepare func(context.Context, *testing.T) (context.Context, error)
```
shashwathi, line 58: stick to one pattern either (context.Context, testing.T) or (testing.T, context.Context)

pkg/cli-runtime/testing/validatable_table.go
```diff
@@ -60,24 +72,31 @@ func (ts ValidatableTestSuite) Run(t *testing.T) {
...
 }
 
-func (tc ValidatableTestCase) Run(t *testing.T) {
+func (tc ValidatableTestCase) Run(t *testing.T, ctx context.Context) {
 	t.Run(tc.Name, func(t *testing.T) {
 		if tc.Skip {
 			t.SkipNow()
 		}
 
-		errs := tc.Validatable.Validate(context.Background())
```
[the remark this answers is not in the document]
shashwathi: +1

pkg/cli-runtime/testing/validatable_table.go
```diff
@@ -98,5 +117,12 @@ func (tc ValidatableTestCase) Run(t *testing.T) {
 		if !tc.ShouldValidate && len(tc.ExpectFieldErrors) == 0 {
 			t.Error("one of ShouldValidate=true or ExpectFieldErrors is required")
 		}
+
+		if tc.CleanUp != nil {
+			var err error
+			if err = tc.CleanUp(ctx, t); err != nil {
+				t.Errorf("unexpected error in clean up: %v", err)
+			}
```
shashwathi, line 125:
```suggestion
			if err := tc.CleanUp(ctx, t); err != nil {
				t.Errorf("unexpected error in clean up: %v", err)
			}
```

pkg/commands/workload_apply_test.go
```diff
@@ -3828,6 +3926,1336 @@ Skipping workload %q`,
 				clitesting.ToInteractTerminal("❓ Really update the workload %q? [yN]: m", workloadName),
 				clitesting.ToInteractTerminal("❓ Really update the workload %q? [yN]: n", workloadName), workloadName),
 		},
+		{
+			Name: "update - replace update strategy",
+			Args: []string{flags.FilePathFlagName, "testdata/replace-update-strategy/all-fields-workload.yaml",
+				flags.TypeFlagName, "my-type",
+				flags.UpdateStrategyFlagName, commands.ReplaceUpdateStrategy, flags.YesFlagName},
```
shashwathi, line 3933: pass flag value in string
```suggestion
				flags.UpdateStrategyFlagName, "replace", flags.YesFlagName},
```

pkg/cli-runtime/testing/validatable_table.go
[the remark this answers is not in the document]
warango4: so when I was trying to create the validatable cases for this, every time the case got to the this part 

```
if opts.UpdateStrategy != "" && cli.CommandFromContext(ctx).Flags().Changed(cli.StripDash(flags.UpdateStrategyFlagName))
```

Since there was just an empty context, there were no flags and I was getting a `nil` error . I needed to add the context in order to set the command in the test during the `preprare` stage

pkg/cli-runtime/testing/validatable_table.go
```diff
@@ -60,24 +72,31 @@ func (ts ValidatableTestSuite) Run(t *testing.T) {
...
 	t.Run(tc.Name, func(t *testing.T) {
 		if tc.Skip {
 			t.SkipNow()
 		}
 
-		errs := tc.Validatable.Validate(context.Background())
+		if tc.Prepare != nil {
+			var err error
+			if ctx, err = tc.Prepare(ctx, t); err != nil {
+				t.Errorf("unexpected error in prepare: %v", err)
```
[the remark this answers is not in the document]
warango4: ok, changed

pkg/cli-runtime/testing/validatable_table.go
warango4: ok, changed

pkg/commands/workload_apply_test.go
warango4: ok, changed

pkg/commands/workload_apply_test.go
```diff
@@ -3828,6 +3926,1336 @@ Skipping workload %q`,
...
+							}, {
+								Name:  "services",
+								Value: apiextensionsv1.JSON{Raw: []byte(`[{"image":"mysql:5.7","name":"mysql"},{"image":"postgres:9.6","name":"postgres"}]`)},
+							},
+						},
+					},
+				},
+			},
+			ExpectOutput: `
+❗ WARNING: Configuration file update strategy is changing. By default, provided configuration files will replace rather than merge existing configuration. The change will take place in the January 2024 TAP release (use "--update-strategy" to control strategy explicitly).
```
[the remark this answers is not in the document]
warango4: ok, added

testing/e2e/workload_test.go
warango4: discussed with PM and showed both approaches to add annotations are working correctly

warango4:
@shaheerkootteeri 

> Do we need any Emojis for the message "workload is unchanged ?" is that message more occurring than before after the introduction of this update-strategy image

I think both of the questions you posted with screenshots are more of a design question. We need to talk to @heyjcollins or @danfein to check what they think about it

pkg/cli-runtime/validation/enum.go
```diff
@@ -35,7 +35,7 @@ func Enum(input string, field string, validOptions []string) FieldErrors {
 
 func EnumInvalidValue(input interface{}, field string, validOptions []string) FieldErrors {
 	return FieldErrors{
-		k8sfield.Invalid(k8sfield.NewPath(field), input, fmt.Sprintf("Supported formats are %v", strings.Join(validOptions, ", "))),
+		k8sfield.Invalid(k8sfield.NewPath(field), input, fmt.Sprintf("supported values are %v", strings.Join(validOptions, ", "))),
```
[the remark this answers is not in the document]
warango4: This message was introduced when we were adding `json` and `yaml` as valid formats to output workload. The message should be more generic, not use the word `formats` because it won't make sense for something like this new flag, where values are `replace` and `merge`. That's why I changed it but we can double check with design

pkg/commands/testdata/replace-update-strategy/replace-annotations.yaml
warango4: Conclusion is: both fields are allowed to be set in `yaml` file, and there are test cases covering both scenarios.

docs/commands-details/workload_create_update_apply.md
```diff
@@ -398,36 +398,23 @@ Create workload:
...
-      22 + |        tag: tap-1.1
-      23 + |      url: https://github.com/sample-accelerators/spring-petclinic
+      1 + |---
+      2 + |apiVersion: carto.run/v1alpha1
+      3 + |kind: Workload
+      4 + |metadata:
+      5 + |  labels:
+      6 + |    apps.tanzu.vmware.com/workload-type: web
+      7 + |  name: spring-pet-clinic
+      8 + |  namespace: default
```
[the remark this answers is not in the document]
warango4: yes, we have that asumption in all our examples/docs that do not specifically mention another `namespace`

pkg/cli-runtime/testing/validatable_table.go
warango4: ok, updated

pkg/cli-runtime/validation/enum.go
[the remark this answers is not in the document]
warango4: @heyjcollins

pkg/cli-runtime/testing/validatable_table.go
shashwathi: I cant see the change yet. Not sure if I am missing something

pkg/commands/workload_apply.go
```diff
@@ -234,6 +273,10 @@ Workload configuration options include:
 
 	// Define common flags
 	opts.DefineFlags(ctx, c, cmd)
+	cmd.Flags().StringVar(&opts.UpdateStrategy, cli.StripDash(flags.UpdateStrategyFlagName), "merge", "specify configuration file update strategy (supported strategies: merge, replace)")
```
shashwathi, line 276: instead of `merge` string, use the const you have declared

shashwathi: merge, replace -> use const as well for this

pkg/commands/workload_apply.go
```diff
@@ -234,6 +273,10 @@ Workload configuration options include:
...
+	cmd.RegisterFlagCompletionFunc(cli.StripDash(flags.UpdateStrategyFlagName), func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return []string{"replace", "merge"}, cobra.ShellCompDirectiveNoFileComp
```
shashwathi, line 278: use the const declared as well

pkg/cli-runtime/testing/validatable_table.go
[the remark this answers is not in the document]
shaheerkootteeri: if you can pass a fresh context to prepare and use that for validate as well as cleanup for each test case the error shouldn't happen. If we use the same context across all the test cases under a suite, wouldnt it hinder test cases that doesnt have a cleanup ?

pkg/commands/workload_apply.go
```diff
@@ -49,8 +52,23 @@ var (
 	_ cli.DryRunable         = (*WorkloadApplyOptions)(nil)
 )
 
+const (
+	MergeUpdateStrategy   = "merge"
+	ReplaceUpdateStrategy = "replace"
```
shaheerkootteeri, line 57: If no other files are using these, can it be private constants

pkg/commands/testdata/replace-update-strategy/replace-subpath.yaml
```diff
@@ -0,0 +1,28 @@
...
+  labels:
+    app.kubernetes.io/part-of: spring-petclinic
+    apps.tanzu.vmware.com/workload-type: web
+spec:
+  source:
+    subPath: ./app
+    git:
+      url: https://github.com/sample-accelerators/spring-petclinic
+      ref:
+        tag: tap-1.1
```
shashwathi, line 28: add a new line at the end of file. same for all new yaml files added in this PR

pkg/commands/workload_apply.go
warango4: ok, changed

pkg/cli-runtime/validation/enum.go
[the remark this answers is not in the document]
heyjcollins: Using `values` rather than `formats` is a sound decision.
Good call @warango4

pkg/commands/workload_apply.go
warango4: changed as well

pkg/commands/workload_apply.go
warango4: changed too

pkg/commands/testdata/replace-update-strategy/replace-subpath.yaml
warango4: lines added

shaheerkootteeri (approved):
LGTM

pkg/cli-runtime/testing/validatable_table.go
```diff
@@ -60,24 +72,31 @@ func (ts ValidatableTestSuite) Run(t *testing.T) {
...
 	t.Run(tc.Name, func(t *testing.T) {
 		if tc.Skip {
 			t.SkipNow()
 		}
 
-		errs := tc.Validatable.Validate(context.Background())
+		if tc.Prepare != nil {
+			var err error
+			if ctx, err = tc.Prepare(t, ctx); err != nil {
+				t.Errorf("error during prepare %v", err)
```
shashwathi, line 95: stick to consistent error msgs. 
```sh
t.Errorf("unexpected error in _func_name_ : %v", err)
```

testing/e2e/workload_test.go
```diff
@@ -516,6 +517,145 @@ func TestCreateFromGitWithAnnotations(t *testing.T) {
 				return nil
 			},
 		},
+		{
+			Name:         "Do not update the created workload with `merge` update strategy",
```
shashwathi, line 521: name of the test in confusing. `create workload with merge strategy`

pkg/commands/workload_apply.go
```diff
@@ -234,6 +273,10 @@ Workload configuration options include:
 
 	// Define common flags
 	opts.DefineFlags(ctx, c, cmd)
+	cmd.Flags().StringVar(&opts.UpdateStrategy, cli.StripDash(flags.UpdateStrategyFlagName), mergeUpdateStrategy, "specify configuration file update strategy (supported strategies: merge, replace)")
```
shashwathi, line 276: (supported strategies: merge, replace) -> use const here as well.

pkg/commands/workload_apply.go
```diff
@@ -101,20 +120,40 @@ func (opts *WorkloadApplyOptions) Exec(ctx context.Context, c *cli.Config) error
...
 	}
 
-	workload.Merge(fileWorkload)
+	if opts.UpdateStrategy == replaceUpdateStrategy {
+		// assign all the file workload fields to the workload in the cluster
+		workload = fileWorkload
+
+		// if there is a workload in the cluster with all metadata populated
+		// re assign the system populated fields so we won't find an error because of some missing fields
+		if currentWorkload != nil && !reflect.DeepEqual(currentWorkload.ObjectMeta, (metav1.ObjectMeta{})) {
```
shashwathi, line 142: nit: create a helper func to move this logic and add unit tests for that

pkg/cli-runtime/testing/validatable_table.go
```diff
@@ -60,24 +72,31 @@ func (ts ValidatableTestSuite) Run(t *testing.T) {
 	}
 	if len(focused) != 0 {
 		for _, tc := range focused {
-			tc.Run(t)
+			tc.Run(t, ctx)
```
shashwathi, line 75: To @shaheerkootteeri  earlier point, context should not shared across all tests. Update func signature to `Run(t)`

testing/e2e/workload_test.go
warango4: this workload is already created, and I created a file to see if it would delete some of the fields

pkg/cli-runtime/testing/validatable_table.go
warango4: then how do i create tests for validate?

> so when I was trying to create the validatable cases for this, every time the case got to the this part
> 
> ```
> if opts.UpdateStrategy != "" && cli.CommandFromContext(ctx).Flags().Changed(cli.StripDash(flags.UpdateStrategyFlagName))
> ```
> 
> Since there was just an empty context, there were no flags and I was getting a `nil` error . I needed to add the context in order to set the command in the test during the `preprare` stage

shashwathi: ctx := context.Background()
create this in the Run func and pass it to validatable

testing/e2e/workload_test.go
shashwathi: Then use Update instead of create

pkg/cli-runtime/testing/validatable_table.go
warango4: But that's what I'm currently doing

warango4: Please [check](https://github.com/vmware-tanzu/apps-cli-plugin/blob/6c161dd35c2388739993cea98a767a91c768bb40/pkg/cli-runtime/testing/validatable_table.go#L66)

pkg/commands/workload_apply.go
warango4: I didn't do it as a separate func because there is no need to have a public func for that logic since it's only being used in apply and it's not likely to be used anywhere else.

pkg/commands/workload_apply.go
warango4: ok, done

pkg/cli-runtime/testing/validatable_table.go
```diff
@@ -60,24 +72,31 @@ func (ts ValidatableTestSuite) Run(t *testing.T) {
...
 	}
 
 	for _, tc := range ts {
-		tc.Run(t)
+		tc.Run(t, ctx)
 	}
 }
 
-func (tc ValidatableTestCase) Run(t *testing.T) {
+func (tc ValidatableTestCase) Run(t *testing.T, ctx context.Context) {
```
shashwathi, line 86:
```suggestion
func (tc ValidatableTestCase) Run(t *testing.T) {
   ctx := context.Background()
```

warango4: ok, updated

pkg/cli-runtime/testing/validatable_table.go
warango4: updated as mentioned in another comment

pkg/commands/workload_apply.go
warango4: ok, separated logic and created unit tests

testing/e2e/workload_test.go
warango4: renamed

pkg/cli-runtime/testing/validatable_table.go
warango4: https://github.com/vmware-tanzu/apps-cli-plugin/pull/402#discussion_r1025759155

warango4: I sticked to the pattern mentioned in the previous comment

pkg/cli-runtime/testing/validatable_table.go
```diff
@@ -98,5 +117,11 @@ func (tc ValidatableTestCase) Run(t *testing.T) {
 		if !tc.ShouldValidate && len(tc.ExpectFieldErrors) == 0 {
 			t.Error("one of ShouldValidate=true or ExpectFieldErrors is required")
 		}
+
+		if tc.CleanUp != nil {
+			if err := tc.CleanUp(t, ctx); err != nil {
+				t.Errorf("unexpected error in clean up: %v", err)
+			}
```
shashwathi, line 124: update the above error msg with `t.Errorf("error during cleanup %v", err)`

warango4: updated

shashwathi (approved):

shaheerkootteeri (approved):

PR #430, opened 2022-12-02
Updates Make file as well as CI for unit test

shaheerkootteeri:
Pull request

Updates the CI as well as MAKE file to remove repeating tasks.

Fixes #

Validated all the make commands in local cluster as well as ran CI successfully.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,26 +46,33 @@ jobs:
         tanzu plugin install builder --local ${TANZU_HOME}/admin-plugins
         tanzu plugin install test --local ${TANZU_HOME}/admin-plugins
         tanzu plugin list
-    - name: Test
-      run: make test
     - name: Scan Inclusive Terminology
       uses: get-woke/woke-action@v0
       with: 
         fail-on-error: true
         woke-args: -c https://via.vmw.com/its-woke-rules
+    - name: Test
+      run: make test
     - name: Build Version
       run: echo "BUILD_VERSION=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
       if: startsWith(github.ref, 'refs/tags/')
     - name: Build
-      run: |
-        make publish
+      run: make build
+    - name: Publish
+      run: make publish
     - uses: actions/upload-artifact@v3
       with:
         name: tanzu-apps-plugin.tar.gz
         path: tanzu-apps-plugin*.tar.gz
         retention-days: 1
     - name: Install
-      run: make install
+      run: |
+        if [[ -z ${BUILD_VERSION} ]]; then      
+          BUILD_SHA_SHORT=$(git rev-parse --short HEAD)
+          BUILD_VERSION=$(cat APPS_PLUGIN_VERSION)+dev-$BUILD_SHA_SHORT         
+        fi
+        echo $BUILD_VERSION
+        tanzu plugin install apps --version $BUILD_VERSION --local ./artifacts/published/linux-amd64
     - name: Verify docs
       run: |
         rm -rf docs/command-reference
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -156,15 +163,14 @@ jobs:
...
-      run: |
-        make publish
+      run: make build
+    - name: Publish
+      run: make publish
     - name: Install
-      run: make install
-    - name: Verify docs
       run: |
-        rm -rf docs/command-reference
```
shashwathi, line 165: is this added somewhere else? this is a crucial step in CI.

.github/workflows/ci.yaml
```diff
@@ -46,26 +46,30 @@ jobs:
...
+      run: make publish
     - uses: actions/upload-artifact@v3
       with:
         name: tanzu-apps-plugin.tar.gz
         path: tanzu-apps-plugin*.tar.gz
         retention-days: 1
     - name: Install
-      run: make install
+      run: |
+        BUILD_SHA_SHORT=$(git rev-parse --short HEAD)
```
shashwathi, line 70: For normal CI run(without tags) , this flow works because `BUILD_VERSION` in previous step is not set. When tag is cut, `BUILD_VERSION` will be set so how does installing plugin which is built with one version and running command with another version succeed?

shaheerkootteeri: Updated commands to validate if BUILD_VERSION is preset before reassigning the version.

anibmurthy (dismissed):
lgtm

Makefile
```diff
@@ -37,21 +37,18 @@ BUILD_JOBS := $(addprefix build-,${ENVS})
...
-install: publish-local ## Install the plugin binaries to the local machine
+install: test## Install the plugin binaries to the local machine
 	@# TODO avoid deleting an existing plugin once in place reinstalls are working again
 	@tanzu plugin delete apps > /dev/null 2>&1 || true
-	tanzu plugin install apps --version $(BUILD_VERSION) --local $(TANZU_PLUGIN_PUBLISH_PATH)/${GOHOSTOS}-${GOHOSTARCH}
-
-.PHONY: build-local
-build-local: prepare test ## Build the plugin binaries locally for the host architecture
-	echo BUILD_VERSION: $(BUILD_VERSION)
 	tanzu builder cli compile --version $(BUILD_VERSION) --ldflags "$(LD_FLAGS)" --path ./cmd/plugin --target local --artifacts ${ARTIFACTS_DIR}/${GOHOSTOS}/${GOHOSTARCH}/cli
```
odinnordico, line 46: is it possible here to call `build-${GOHOSTOS}-${GOHOSTARCH}`? instead writing the command again?

Makefile
```diff
@@ -37,21 +37,18 @@ BUILD_JOBS := $(addprefix build-,${ENVS})
...
+	tanzu builder publish --type local --plugins "apps" --version $(BUILD_VERSION) --os-arch "${GOHOSTOS}-${GOHOSTARCH}" --local-output-discovery-dir "$(TANZU_PLUGIN_PUBLISH_PATH)/${GOHOSTOS}-${GOHOSTARCH}/discovery/standalone" --local-output-distribution-dir "$(TANZU_PLUGIN_PUBLISH_PATH)/${GOHOSTOS}-${GOHOSTARCH}/distribution" --input-artifact-dir $(ARTIFACTS_DIR)
```
odinnordico, line 47: is it possible here to call `publish-${GOHOSTOS}-${GOHOSTARCH}`? instead writing the command again?

.github/workflows/ci.yaml
```diff
@@ -46,26 +46,36 @@ jobs:
...
       with:
         name: tanzu-apps-plugin.tar.gz
         path: tanzu-apps-plugin*.tar.gz
         retention-days: 1
     - name: Install
-      run: make install
+      run: |
+        if [[ ${BUILD_VERSION} ]]; then
+          echo $BUILD_VERSION
+          tanzu plugin install apps --version $BUILD_VERSION --local ./artifacts/published/linux-amd64
```
odinnordico, line 72: you can leave the command out of the `if` as it is the same fot the two branches (`if`/`else`)

.github/workflows/ci.yaml
```diff
@@ -156,16 +168,20 @@ jobs:
...
-      run: make install
-    - name: Verify docs
       run: |
-        rm -rf docs/command-reference
-        tanzu apps docs --directory docs/command-reference
-        git diff --exit-code docs
-      continue-on-error: true
+        if [[ ${BUILD_VERSION} ]]; then
+          echo $BUILD_VERSION
+          tanzu plugin install apps --version $BUILD_VERSION --local ./artifacts/published/darwin-amd64
```
odinnordico, line 178: you can leave the command out of the `if` as it is the same fot the two branches (`if`/`else`)

Makefile
shaheerkootteeri: builder command in local install has --target set as local , while the build -% PHONY has the target set as --target ${OS}_${ARCH}

Makefile
shaheerkootteeri: We can call the `publish-${GOHOSTOS}-${GOHOSTARCH}` however it would create a tar file that is not necessary for install. We wanted to avoid creating a tar while calling make install.

Makefile
odinnordico: `local` can be interpreted as a shortcut for `runtime.GOOS, runtime.GOARCH` but I am good with leaving it a command to keep consistency in the `make` target

.github/workflows/ci.yaml
shaheerkootteeri: Updated the code to assign build version only if its not set, and tanzu plugin command moved outside

.github/workflows/ci.yaml
shaheerkootteeri: Updated the code to assign build version only if its not set, and tanzu plugin command moved outside

odinnordico (approved):
LGTM

warango4 (approved):

anibmurthy (approved):

PR #460, opened 2023-01-26
Add --output flag to create/apply

### LINKED ISSUE #392: `workload create/apply` supports `--output yaml/json`
### Description of problem

When scripting the creation/update of workloads today, users must execute the `tanzu apps workload create/apply`, wait 'till the workload has been created, then execute a 2nd `tanzu apps workload get app-name -oyaml`.

Prior art for supporting this type of interaction:
* `kubectl apply -f file.yaml -oyaml`
* `imgpkg push --json`

### Proposed solution

Supporting `-o yaml/json` will simplify this interaction and eliminate this extra toil

* [ ] `-o yaml/json` implies `--wait`
* [ ] tanzu apps workload create/apply help is updated
  * `-o, --output string    output the Workload formatted. Supported formats: "json", "yaml", "yml"`
* [ ] new flag support added to release notes
* [ ] example utilization is provided in documentation

warango4:
Pull request

Add `--output` flag to create/apply to retrieve workload once it's ready (mostly for scripting purposes)

Fixes #392 

### DIFF
--- a/pkg/commands/workload.go
+++ b/pkg/commands/workload.go
@@ -46,6 +46,8 @@ import (
 	"github.com/vmware-tanzu/apps-cli-plugin/pkg/cli-runtime/parsers"
 	cliprinter "github.com/vmware-tanzu/apps-cli-plugin/pkg/cli-runtime/printer"
 	"github.com/vmware-tanzu/apps-cli-plugin/pkg/cli-runtime/validation"
+	"github.com/vmware-tanzu/apps-cli-plugin/pkg/cli-runtime/wait"
+	"github.com/vmware-tanzu/apps-cli-plugin/pkg/cli-runtime/watch"
 	"github.com/vmware-tanzu/apps-cli-plugin/pkg/completion"
 	"github.com/vmware-tanzu/apps-cli-plugin/pkg/flags"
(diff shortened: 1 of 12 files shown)


pkg/commands/workload.go
```diff
@@ -494,6 +513,56 @@ func (opts *WorkloadOptions) PublishLocalSource(ctx context.Context, c *cli.Conf
 	return okToPush, nil
 }
 
+func (opts *WorkloadOptions) PublishLocalSourceWithoutPrompts(ctx context.Context, c *cli.Config, currentWorkload, workload *cartov1alpha1.Workload) error {
```
anibmurthy, line 516: I understand the intention here is to skip printing anything so that `--output` does not include any text. If that is the case, instead of having a separate function which is mostly a replica of PushLocalSource have a utility function with an indicator to either print or silent return. That way we don't have to maintain two implementations.

odinnordico (approved):

anibmurthy (approved):
lgtm

pkg/commands/workload.go
warango4: Thanks Ani, 

I changed it to be one function that would check in every step if there should be any prompts

PR #470, opened 2023-02-14
Add client side for LSP

### LINKED ISSUE #364: Enable CLI to connect to source registry proxy within the platform
### Description of problem

If `tanzu` platform is deployed with a registry within the cluster, CLI cannot upload source code to that registry as it needs to pass `source-image` for local source code flow.

### Proposed solution

Make  `--source-image` flag optional. This will enable users to use registry that's deployed within the cluster.  CLI can initiate HTTP client that communicates with service (that's is targeting the registry ) securely.
 
```gherkin
Given workload create/apply command 
When command is passed without source-image
Then CLI connects with k8s service (registry name/ns) to check for local source presence.
```

Things to do 

- [ ] Make source-image flag optional
- [ ] If source image is not provided then build HTTP client to connect to source registry
   - [ ] This connection should be secure and to avoid using cert, use kubernetes client to generate HTTP request with RBAC permissions of the current context
   - [ ] If connection to registry fails, then print error `No source registry found.`
- [ ] If user provides source image flag then use that for uploading source registry
- [ ] Add e2e test (this requires setting up docker registry in a container and k8s service)

warango4:
Pull request

Add client side for Local Source Proxy, with unit test and adding new annotations for workloads which will use this kind of registry

Fixes #364 

- Removed tests which check for `source-image` as mandatory flag when using local source.
- Added new unit testing creating fake RoundTrip and local registry to push to.
- Unit tests were updated to use source image and no source image to analyze different scenarios.
- Added validation in e2e test to check that annotation is not being added unless source code is going to be pushed to local registry.

There is no unit test to check annotation inside `workload create/apply` commands, because we currently do not have testing for the complete flow between these commands and PublishLocalSource() func (in unit). 
We do test this flow in e2e but these kind ot tests are still a work in progress and will be added in a future PR when LSP can be installed as part of the CI flow.


pkg/source/fake/fake_wrapper.go
```diff
@@ -0,0 +1,47 @@
...
+		Body:       nil,
+		Request:    req,
+	}, nil
+}
+
+func GetFakeWrapper() source.Wrapper {
+	return source.Wrapper{
+		Client: &http.Client{Transport: nilBodyRoundTripper{}},
+		URL: &url.URL{
+			Scheme: "http",
```
odinnordico, line 41: does this URL need to use `google`? if no, please use some vmw url or any non existing one

pkg/commands/workload.go
```diff
@@ -181,9 +183,6 @@ func (opts *WorkloadOptions) Validate(ctx context.Context) validation.FieldError
 	}
 
 	if opts.RegistryPassword != "" || opts.RegistryUsername != "" || opts.RegistryToken != "" || len(opts.CACertPaths) != 0 {
-		if opts.SourceImage == "" {
-			errs = errs.Also(validation.ErrMissingField(flags.SourceImageFlagName))
-		}
```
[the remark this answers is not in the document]
warango4: yes, I missed this and asked some clarification for it. Thanks for pointing it out.

pkg/commands/workload.go
```diff
@@ -58,8 +58,11 @@ import (
 const (
 	AnnotationReservedKey     = "annotations"
 	MavenOverwrittenNoticeMsg = "Maven configuration flags have overwritten values provided by \"--params-yaml\"."
+	SourceProxyDomain         = "svc.cluster.local"
 )
 
+var LocalImageRepo = fmt.Sprintf("%s.%s.%s", source.SourceProxyService, source.SourceProxyNamespace, SourceProxyDomain)
```
anibmurthy, line 64: *Suggestion:* Per latest plan, we could combine these 3 constants into 1 field. 

We had made it configurable in first version with an intention to support change in lsp endpoint for any reason. However, we moved away from that plan and will have same name/namespaces across releases.

pkg/commands/workload_apply.go
```diff
@@ -158,20 +153,31 @@ func (opts *WorkloadApplyOptions) Exec(ctx context.Context, c *cli.Config) error
...
+	} else if !okToPush {
+		return nil
+	}
+	// merge annotation only when workload is being created, do not add it when updating workload
+	// since user could be updating another field and annotation must not be added
+	if opts.LocalPath != "" && opts.SourceImage == "" && !workloadExists {
+		workload.MergeAnnotations(apis.LocalSourceProxyAnnotationName, workload.Spec.Source.Image)
+	}
+	// if workload is updated from LSP registry to custom or any other registry through source image,
+	// annotation has to be deleted and workload source image needs to be updated to digested based on opts source image
```
shaheerkootteeri, line 169: We would need similar validation in workload update as well ?

testing/e2e/workload_test.go
```diff
@@ -421,6 +426,9 @@ func TestCreateFromGitWithAnnotations(t *testing.T) {
 				if regexProgress.FindString(output) != "" {
 					t.Fatalf("expected progressbar not to be in output %v", output)
 				}
+				if strings.Contains(output, apis.LocalSourceProxyAnnotationName) {
+					t.Fatalf("local source proxy annotation %q not expected in output %v", apis.LocalSourceProxyAnnotationName, output)
```
shaheerkootteeri, line 430: Can we also add a test for failure scenarios , when the workload was created using `--source-image` and trying to update without a repo details?

pkg/commands/workload_apply.go
warango4: I don't think we should add validation for `uptate` command since it's being deprecated. My understanding is that we should not keep on supporting this command

testing/e2e/workload_test.go
warango4: @atmandhol if user changes something in the local source and then want to apply but they do not use the source image, should this fail?

atmandhol: No, you only have to provide --source-image flag once. For all subsequent calls, the source-image value is taken from the workload.spec.source.image field

pkg/commands/workload.go
```diff
@@ -450,8 +449,19 @@ func (opts *WorkloadOptions) PublishLocalSource(ctx context.Context, c *cli.Conf
 		return true, nil
 	}
 
+	var taggedImage string
 	var okToPush bool
-	taggedImage := strings.Split(workload.Spec.Source.Image, "@sha")[0]
+
+	isLocal := opts.SourceImage == ""
```
shaheerkootteeri, line 455: for setting islocal true and update the source image url ,we might have to consider if it is creation or updating as well as the annotation.

pkg/source/fake/fake_wrapper.go
warango4: ok, updated to fake URL

pkg/commands/workload.go
```diff
@@ -488,6 +498,18 @@ func (opts *WorkloadOptions) PublishLocalSource(ctx context.Context, c *cli.Conf
 		return false, fmt.Errorf("unsupported file format %q", opts.LocalPath)
 	}
 
+	localTransport := &source.Wrapper{}
+	if isLocal {
+		var err error
+		clientSet := c.GetClientSet()
+		// pass RESTClient as CoreV1 restclient, which will call custom RoundTripper
+		localTransport, err = source.LocalRegistryTransport(ctx, clientSet, c.KubeRestConfig(), clientSet.CoreV1().RESTClient())
```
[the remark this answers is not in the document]
warango4: yes, you're right. Removed the need for clientset

pkg/commands/workload.go
warango4: decided to move the variable to be part of the local registry transport, which contains the other constants. Decided to keep it as variable since the other vars that it's composed from are actually needed in other parts of the code.

testing/e2e/workload_test.go
warango4: updated to reflect this behavior

pkg/commands/workload.go
warango4: I also thought about adding a new method `GetLocalImageRepo()` which would return this variable and that way there's no way the caller function would override this variable, what do you think?

pkg/apis/cartographer/v1alpha1/workload_helpers.go
```diff
@@ -425,6 +425,15 @@ func (w *Workload) MergeAnnotations(key, value string) {
 	w.Annotations[key] = value
 }
 
+func (w *Workload) RemoveAnnotations(key string) {
+	delete(w.Annotations, key)
+}
+
+func (w *Workload) AnnotationExists(key string) bool {
```
anibmurthy, line 432: *nit:* Would you like to call this `IsAnnotationExists` ?

warango4: ok, changed

shaheerkootteeri (approved):
LGTM

anibmurthy (dismissed):
lgtm

Makefile
```diff
@@ -73,7 +73,7 @@ docs: $(GO_SOURCES) ## Generate the plugin documentation
 
 .PHONY: test
 test: prepare## Run tests
-	go test ./... -coverprofile=coverage.txt -covermode=atomic -timeout 1m -race
+	go test ./... -coverprofile=coverage.txt -covermode=atomic -timeout 3m -race
```
anibmurthy, line 76: I see this change as part of #474 watch out for merge conflicts

pkg/commands/workload_apply.go
```diff
@@ -158,20 +152,22 @@ func (opts *WorkloadApplyOptions) Exec(ctx context.Context, c *cli.Config) error
...
 	}
 
+	workloadExists := currentWorkload != nil
+
+	if okToPush, err := opts.PublishLocalSource(ctx, c, currentWorkload, workload, shouldPrint); err != nil {
+		return err
+	} else if !okToPush {
+		return nil
+	}
+	opts.ManageLocalSourceProxyAnnotation(currentWorkload, workload)
```
anibmurthy, line 162: *Question:* Could this be moved to `workload` (common) instead of calling this in 3 different functions?

warango4: it didn't seem as the best decision for me, because the other part where we could call this func is inside `PublishLocalSource()` and, for me, this bit of the annotation is not a concern for that func, so I decided to add it in exec despite it requires a bit of repetition.

Makefile
warango4: yes, will have it in mind

anibmurthy (approved):

odinnordico (approved):

PR #498, opened 2023-03-03
Default workload type to web

### LINKED ISSUE #469: default workload type set to `web` for `workload create/apply`
### Description of problem

Creating a workload via the apps plugin requires familiarity with a significant number of `create/apply` flags.
Providing sane/likely defaults for some of those flags will reduce onboarding friction as well as increasing the day-to-day efficiency of regular users.

### Proposed solution

Provide a default value of `web` for the `--type` flag.

```gherkin
Given I don't provide `--type` flag/value
When I run `tanzu apps workload create/apply...`
Then my workload of type `web` is created successfully
```

warango4:
Pull request

Defaulting workload `type` to `web` when workload is going to be created.

Fixes #469 

- Updated unit test
- Updated e2e test to guarantee that worklodas are being created defaulting `type` to `web` when `--type` flag is not used.


pkg/commands/workload.go
```diff
@@ -807,9 +809,9 @@ func (opts *WorkloadOptions) DefineFlags(ctx context.Context, c *cli.Config, cmd
 	cli.NamespaceFlag(ctx, cmd, c, &opts.Namespace)
 	cmd.Flags().StringVarP(&opts.FilePath, cli.StripDash(flags.FilePathFlagName), "f", "", "`file path` containing the description of a single workload, other flags are layered on top of this resource. Use value \"-\" to read from stdin")
 	cmd.Flags().StringVarP(&opts.App, cli.StripDash(flags.AppFlagName), "a", "", "application `name` the workload is a part of")
-	cmd.Flags().StringVarP(&opts.Type, cli.StripDash(flags.TypeFlagName), "t", "", "distinguish workload `type`")
+	cmd.Flags().StringVarP(&opts.Type, cli.StripDash(flags.TypeFlagName), "t", "web", "distinguish workload `type`")
```
odinnordico, line 812: why dont you use `WebTypeReservedKey` here?

warango4: changed to use the reserved key

pkg/commands/workload_apply_test.go
```diff
@@ -265,7 +274,9 @@ To get status: "tanzu apps workload get my-workload"
 					ObjectMeta: metav1.ObjectMeta{
 						Namespace: defaultNamespace,
 						Name:      workloadName,
-						Labels:    map[string]string{},
+						Labels: map[string]string{
+							apis.WorkloadTypeLabelName: "web",
```
shaheerkootteeri, line 278: Is it possible to validate that , update to an existing workload doesn't add the type or modify the type via apply or update?

pkg/commands/workload_apply_test.go
```diff
@@ -3745,10 +3906,14 @@ To get status: "tanzu apps workload get spring-petclinic"
 `,
 		},
 		{
-			Name: "update git fields",
+			Name: "update - do not default type in existing workload",
```
warango4, line 3909: @shaheerkootteeri added new test case to avoid adding `type` in existing workload

pkg/commands/workload_apply_test.go
warango4: added test case, pointed in a different comment

shaheerkootteeri (approved):
LGTM

PR #516, opened 2023-03-21
Add autocompletion for type flag

### LINKED ISSUE #360: Tab completion doesn't work for `--type` flag values

### Describe the bug
Tab completion works for flag values, but doesn't work for `--type`

### Expected behavior
When I hit <TAB> after typing `tanzu apps workload create/apply APP-NAME --type`
I should see the value(s) for all of the workload types that are supported on the target cluster

### Steps to Reproduce
1. target a cluster with supply chain(s) that support multiple workload types
2. type `tanzu apps workload create/apply APP-NAME --type ` (include a space after "--type"
3. hit <TAB>

```sh
``` 
version: v0.27.0-dev
buildDate: 2022-09-13
sha: 0b58fbea
v0.9.0+dev-baf0aa02

This issue occurs regardless of which supply chain versions are running on the target cluster

warango4:
Pull request

Add autocompletion func for `--type` flag. 

Fixes #360 

- Modified OOTB supply chain to have different selectors besides `web`, `server`, `worker` and showed also the custom ones.

### DIFF
--- a/pkg/commands/workload.go
+++ b/pkg/commands/workload.go
@@ -36,6 +36,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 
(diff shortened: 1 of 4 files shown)


pkg/commands/workload.go
```diff
@@ -792,13 +794,35 @@ func (opts *WorkloadOptions) GetTailWorkers(c *cli.Config, workload *cartov1alph
 	return workers
 }
 
+func getClusterSupplyChainTypeSelectors(fields []metav1.LabelSelectorRequirement) []string {
+	var values []string
+	for _, v := range fields {
+		if v.Key == apis.WorkloadTypeLabelName && len(v.Values) != 0 {
```
odinnordico, line 800: `len` check is not required, see https://go.dev/play/p/7tJe1k_QO_v

warango4: ok, removed `len` validation

odinnordico (approved):

pkg/commands/workload.go
```diff
@@ -41,6 +41,8 @@ import (
 
 	"github.com/vmware-tanzu/carvel-imgpkg/pkg/imgpkg/registry"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
shaheerkootteeri, line 44: Please group dependencies as below 
Format:

```

Standard library packages

External packages 

Local source code packages

```

warango4: Updated

shaheerkootteeri (approved):

odinnordico (approved):

PR #547, opened 2023-05-05
fix: Use url decode in the output message from lsp

### LINKED ISSUE #540: lsp health check error message formatting
### Description of problem
`lsp health` error messages displayed are url encoded that replaces special characters with an encoded character impacting readability.
<Please describe the problem your feature request is trying to solve>

### Proposed solution
Use url decode to show raw url string when available.

```gherkin
Given `tanzu apps lsp health` call is made
When An error is returned in the response that is printed under error message
Then Error message must be readable without any encoded characters
```

#### current:
```
user_has_permission: true
reachable: true
upstream_authenticated: false
overall_health: false
message: |-
  Local source proxy was unable to authenticate against the target registry.
  Errors:
  - GET https://gcr.io/v2/token?scope=repository%3Aadhol-playground%2Flsp-source%3Apull%2Cpush&service=gcr.io: UNAUTHORIZED: Not Authorized.
```

#### proposed:
```
user_has_permission: true
reachable: true
upstream_authenticated: false
overall_health: false
message: |-
  Local source proxy was unable to authenticate against the target registry.
  Errors:
  - GET https://gcr.io/v2/token?scope=repository:adhol-playground/lsp-source:pull,push&service=gcr.io: UNAUTHORIZED: Not Authorized.
```

odinnordico:
Pull request

When an error occurs calling LSP or when LSP returns a message which contains url encoded text (% followed by two hexadecimal digits), it is shown as it comes.
This PR try to use a URL Decoding strategy to show better the message from LSP, if the decoding strategy fails it shows the message as is

Fixes #540

### DIFF
--- a/pkg/commands/lsp/lsp.go
+++ b/pkg/commands/lsp/lsp.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"strconv"
 
 	"github.com/vmware-tanzu/apps-cli-plugin/pkg/apis/lsp"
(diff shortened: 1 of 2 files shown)


shaheerkootteeri (approved):
LGTM

warango4 (approved):

PR #568, opened 2023-05-17
Adds validations to check multiple sources via flags on workload…

### LINKED ISSUE #549: Workload create/apply lacks validation to prevent inclusion of multiple source types

### Describe the bug
It's possible to create a workload today that includes multiple sources:
1. `--maven*`
2. `--git*`
3. `--image`
4. `-local-path` with/without `--source-image`

### Expected behavior
The four sources should be mutually exclusive and there should be client-side validation prevents submitting more than 1.

### Steps to Reproduce
Should not be able to submit the following workload apply/create command:
```
tanzu apps workload create all-source --git-repo foo.com/bar --git-branch main --image foo --local-path . --source-image gcr.io/pivotal-knat
ive/all-source --maven-group mg --maven-type mt --maven-artifact ma --maven-version mv
```

### Screenshots
Screenshot 2023-05-08 at 12 22 53 PM

shaheerkootteeri:
… create or update

Pull request

PR adds validations to the flags on apps cli command to display error when multiple sources are provided.

Fixes #549 

Added unit test cases to validate the multiple sources scenario. Validated the changes against my local cluster also confirmed integration tests are running successfully. 

PR also include the auto generated (codegen/ diegen) deep copy as well as dies files.


pkg/commands/workload.go
```diff
@@ -64,6 +65,8 @@ const (
 	AnnotationReservedKey     = "annotations"
 	MavenOverwrittenNoticeMsg = "Maven configuration flags have overwritten values provided by \"--params-yaml\"."
 	WebTypeReservedKey        = "web"
+	MavenFlagWildcard         = "--maven*"
+	LocalPathAndSource        = "--local-path with/without --source-image"
```
warango4, line 69: I think it's important to highlight that source image can come without local path, so that's something we should have in mind.

pkg/commands/workload.go
```diff
@@ -228,6 +261,12 @@ func (opts *WorkloadOptions) LoadDefaults(c *cli.Config) {
 	opts.ExcludePathFile = c.TanzuIgnoreFile
 }
 
+func ErrMultipleSourceOf(names ...string) validation.FieldErrors {
```
warango4, line 264: I know that this is something specific for workload, but we have some other files /folders  where we record the actual error types, so I'd move this there.

warango4: Also, should this have its own test function?

pkg/commands/workload.go
shaheerkootteeri: added a comment to address the recommended

pkg/commands/workload.go
shaheerkootteeri: I have moved the function under validations and added unit test cases

pkg/commands/workload.go
```diff
@@ -200,6 +205,34 @@ func (opts *WorkloadOptions) Validate(ctx context.Context) validation.FieldError
...
+	}
+	for _, p := range opts.ParamsYaml {
+		kv := parsers.DeletableKeyValue(p)
+		if len(kv) != 1 {
+			if kv[0] == cartov1alpha1.WorkloadMavenParam {
+				mavenSource = true
+				break
+			}
+		}
+	}
```
anibmurthy, line 220: You would want to check this if `mavenSource` was not set to true in above `if` conditional?

shaheerkootteeri: if either of the mavan sources are there in the command, we need to show that message and hence we would be ok even if the value was set as true before.

pkg/commands/workload.go
```diff
@@ -200,6 +205,34 @@ func (opts *WorkloadOptions) Validate(ctx context.Context) validation.FieldError
...
+	if mavenSource {
+		sources = append(sources, MavenFlagWildcard)
+	}
+	if opts.LocalPath != "" || opts.SourceImage != "" {
+		sources = append(sources, LocalPathAndSource)
+	}
```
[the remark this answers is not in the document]
shaheerkootteeri: The error message states local path with/ without source image. We would need to handle scenario where customer gave only --sourec-image flag as well as only --local-path as those are valid scenarios.

pkg/commands/workload.go
anibmurthy: if `mavenSource` is already true, why would you loop again to make it `true`?

anibmurthy (approved):

odinnordico (approved):

PR #580, opened 2023-05-24
fix: Update subpath from DeepCopy instead of pointer

odinnordico:
Pull request

### Which issue(s) this PR fixes

Fixes #565

PR #598, opened 2023-06-13
fix: Revert "Update subpath from DeepCopy instead of pointer"

warango4:
This reverts commit 3cedc297c6eacc816f810980fa01f6dcf41a3408.

Pull request

Revert unexpected breaking change introduced in #580 

N/A


warango4:
When changing from Git source to LSP

image

warango4:
When changing from LSP to source image, subpath stays

image

PR #650, opened 2023-10-12
fix: Add woke rules in project

warango4:
Pull request

Add woke rules yaml file in project to avoid errors when trying to access it URL

N/A


anibmurthy (dismissed):
Let us use a `woke` dir under `hack` to host the rules file.

